PR #4, opened 2022-08-02
Makefile: Remove deppy dependency

oceanc80:
Update the Makefile and remove deppy dependency.
Also updates BundleInstance to BundleDeployment to be consistent with new RukPak naming.

Signed-off-by: Catherine Chan-Tse <cchantse@redhat.com>


timflannagan (approved):
Just taking a quick glance and I think we need to still clean up some of the repository contents here, but we can take that in a follow-up imo.

/lgtm

timflannagan:
Hmm weird.

/approve
/lgtm

PR #6, opened 2022-08-04
Update the Dockerfile to use downstream images

timflannagan:
Update dockerfile and use the rhel-related base images. Run the build Makefile target in the builder image and copy that PO manager executable into the final image layer.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,16 @@
-FROM gcr.io/distroless/static:debug-nonroot
-WORKDIR /
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.12 AS builder
 
-COPY manager manager
-EXPOSE 8080
+WORKDIR /build
+COPY . .
+RUN make build
+
+FROM registry.ci.openshift.org/ocp/4.12:base
+
+COPY --from=builder /build/bin/manager /bin
+USER 1001
+
+LABEL io.k8s.display-name="OpenShift Platform Operator Manager" \
+      io.k8s.description="This is a component of OpenShift Container Platform and manages the lifecycle of platform operators." \
+      io.openshift.tags="openshift" \
+      summary="This is a component of OpenShift Container Platform and manages the lifecycle of platform operators." \
+      maintainer="Odin Team <aos-odin@redhat.com>"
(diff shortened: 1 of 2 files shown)


test/e2e/platform_operators_test.go
```diff
@@ -159,7 +159,7 @@ var _ = Describe("platform operators controller", func() {
 					if err := c.Get(ctx, types.NamespacedName{Name: po.GetName()}, bi); err != nil {
 						return nil, err
 					}
-					if bi.Status.InstalledBundleName == "" {
+					if bi.Status.ActiveBundle == "" {
```
tylerslaton, line 162: Is this supposed to come along?

tylerslaton (commented):
/lgtm

test/e2e/platform_operators_test.go
timflannagan: Yep, but I had sneaked it in. My IDE was complaining about it locally.

exdx: Yes, I think it's ok -- we changed the name of that field upstream and it probably never got updated in the e2e test on the PO side.

Dockerfile
```diff
@@ -1,5 +1,16 @@
...
+FROM registry.ci.openshift.org/ocp/4.12:base
+
+COPY --from=builder /build/bin/manager /bin
+USER 1001
+
+LABEL io.k8s.display-name="OpenShift Platform Operator Manager" \
+      io.k8s.description="This is a component of OpenShift Container Platform and manages the lifecycle of platform operators." \
+      io.openshift.tags="openshift" \
+      summary="This is a component of OpenShift Container Platform and manages the lifecycle of platform operators." \
+      maintainer="Odin Team <aos-odin@redhat.com>"
```
exdx, line 16: nit: can we update this to a more recent email? We could use the olm dev working group email. 

That email is set up from Evan and I'm not sure the current status of who is monitoring it.

exdx (approved):
/lgtm

Dockerfile
timflannagan: Yep, just a copy/paste error. We'll have this issue across multiple repos, so probably best to update it uniformly.

PR #11, opened 2022-08-09
feat: introduce rukpak manifests into source

tylerslaton:
## Summary
This PR adds the Rukpak manifests into the root of Platform Operators so that, in this interim period between it being added into the core payload, it can come shipped with Rukpak as needed. 

There are a lot of changes here but most of them are directly ported from downstream Rukpak. The main file I would keep an eye on is the Makefile as well the repo general structure.

> *Note*: These manifests still require that cert-manager be installed in order to run. We should think if we want cert-manager to be a dependency here as well and introduce it into its installation temporarily or avoid it all together.

## Not handled here, but should consider
1. Automatically syncing the manifests between [downstream rukpak](https://github.com/openshift/operator-framework-rukpak) and here.
2. Adding automation to increment the version of the Rukpak image used.
3. Removal of cert-manager as a dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,9 +63,11 @@ RBAC_LIST = rbac.authorization.k8s.io_v1_clusterrole_platform-operators-manager-
 .PHONY: manifests
 manifests: generate kustomize
 	$(KUSTOMIZE) build config/default -o $(TMP_DIR)/
+	$(KUSTOMIZE) build config/rukpak -o $(TMP_DIR)/rukpak.yaml
 	ls $(TMP_DIR)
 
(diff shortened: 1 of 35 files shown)


timflannagan (approved):
We talked in slack but I think we need to separate out the rukpak.yaml into individual manifests at the minimum. Also, I _think_ we need to update the kustomize namespace prefix and append "openshift-*" to our rukpak system namespace.

timflannagan:
This doesn't affect anything e2e facing right now.

/override ci/prow/e2e

PR #12, opened 2022-08-09
chore: remove crd validator

tylerslaton:
Removing the `crdvalidator` manifests from the downstream as they are not needed.


timflannagan (approved):
Unrelated: Any idea where the rukpak webhook configuration files are stored?

timflannagan:
Same comment as before: these don't affect e2e results yet.

/override ci/prow/e2e

PR #13, opened 2022-08-10
refactor: break rukpak manifests into separated files

tylerslaton:
## Summary
Breaking out the rukpak manifests into their own unique files. This is accomplished by doing a few things:

1. Adding the `rukpak` directory into the `default` directory's `kustomize.yaml` as a base
2. Creating a prefix for all `rukpak` resources to come with `rukpak-`
3. Update the `Makefile` to call `kustomize` once and then move all of the rukpak files into the `manifests` directory
4. Add the generated RBAC to the list for generation.

Again, there are a lot of changes here but the main things to check come down to: 

- File structure
- Name spacing for each resource
- Makefile changes
- Kustomize changes
- Deleted files

## Open questions
- Do we want to bring this upstream as well? This seems like a lot of work when our current unified `rukpak.yaml` seems to work pretty well for the majority of cases there. The main reason it is downstream is for the CVO to consume.


manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: platform-operators-rukpak-webhook-certificate
+  namespace: openshift-platform-operators-system
```
exdx, line 5: nit: if this were the namespace, I think we would drop the `system` suffix to be idiomatic with other openshift namespaces.

Makefile
```diff
@@ -57,28 +57,47 @@ RBAC_LIST = rbac.authorization.k8s.io_v1_clusterrole_platform-operators-manager-
...
-	mv $(TMP_DIR)/rukpak.yaml manifests/0000_50_cluster-platform-operator-manager_00-rukpak.yaml
-	mv $(TMP_DIR)/apiextensions.k8s.io_v1_customresourcedefinition_platformoperators.platform.openshift.io.yaml manifests/0000_50_cluster-platform-operator-manager_00-platformoperator.crd.yaml
-	mv $(TMP_DIR)/v1_namespace_openshift-platform-operators-system.yaml manifests/0000_50_cluster-platform-operator-manager_00-namespace.yaml
-	mv $(TMP_DIR)/v1_serviceaccount_platform-operators-controller-manager.yaml manifests/0000_50_cluster-platform-operator-manager_01-serviceaccount.yaml
-	mv $(TMP_DIR)/v1_service_platform-operators-controller-manager-metrics-service.yaml manifests/0000_50_cluster-platform-operator-manager_02-metricsservice.yaml
-	mv $(TMP_DIR)/apps_v1_deployment_platform-operators-controller-manager.yaml manifests/0000_50_cluster-platform-operator-manager_06-deployment.yaml
+	@# Move all of the rukpak manifests into the manifests folder
+	mv $(TMP_DIR)/apiextensions.k8s.io_v1_customresourcedefinition_bundledeployments.core.rukpak.io.yaml manifests/0000_60_cluster-platform-operator-manager_00-rukpak-bundledeployments.crd.yaml
```
exdx, line 73: should we include `mv` as part of the variable? Something like $MV_TMP_DIR=`mv $(TMP_DIR)`?

config/rukpak/core/resources/rukpak_issuer.yaml
```diff
@@ -2,15 +2,15 @@
...
 spec:
   selfSigned: {}
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: rukpak-ca
-  namespace: rukpak-system
+  name: ca
+  namespace: system
```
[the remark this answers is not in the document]
tylerslaton: So this gets auto-generated from `kustomize` into `openshift-platform-operators-system`. This namespace gets created from [this file](https://github.com/openshift/platform-operators/blob/43c1875dec8baf84af6f8d572d5d60a7011c40a0/manifests/0000_60_cluster-platform-operator-manager_00-namespace.yaml). While it doesn't automatically come with Openshift, we do create it first before other resources that are namespaced.

manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
tylerslaton: This is the right namespace I believe as it is just using the namespace that other platform operators manifests use. That being said, we could look at dropping the `system` suffix to be more idiomatic.

Makefile
tylerslaton: That feels cleaner than it is now, yeah. We should consider potentially making a commit into `operator-sdk` that updates that there as well as this is the default.

timflannagan: Quick note: this repository wasn't bootstrapped using operator-sdk, and the manifests target here is custom/hacky logic built on top of kubebuilder.

config/rukpak/core/resources/rukpak_issuer.yaml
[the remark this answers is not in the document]
timflannagan: imo, we deploy rukpak and PO in separate namespaces for right now so we continue to make progress here. if this is a blocker for payload introduction (for whatever) reason, we can easily consolidate this back to a single namespace. I think my preference here would be a net-new namespace for all of this content, vs. consolidating everything into that openshift-operator-lifecycle-manager namespace.

Makefile
tylerslaton: Ahh, gotcha gotcha. Appreciate that context then. I added the update so our hacky logic becomes a bit cleaner in that case.

manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
timflannagan: I'm not sure if `system` suffix's are less idiomatic now that new components are being bootstrapped with this in mind. I'd be surprised if you fired up a newer 4.x version, and there's no namespaces with this system suffix tbh.

config/rukpak/core/resources/rukpak_issuer.yaml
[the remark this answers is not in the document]
tylerslaton: They're currently both in the same namespace as its easier to consolidate them into PO's namespace. Are we okay with having a net new namespace that contains both Platform Operators and Rukpak?

manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
tylerslaton: Yeah, fair enough. I'll boot up a 4.x cluster just to make sure we're in the green here wrt to openshift idioms.

tylerslaton: So looking at a 4.11 cluster I found no default OpenShift namespaces that had a `-system` suffix.

exdx: Yes, I would recommend dropping `system` and having the namespace be `openshift-platform-operators` to be more idiomatic with other openshift namespaces

exdx:
Looks like there's no kubectl in the container, which causes e2e to fail. Surely we get around this in OLM somehow?

manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
tylerslaton: Let me know how we feel about this. I have some changes standing-by that remove the `-system` suffix. If we want to go down that path I can go ahead and push it up. Additionally, I couldn't find anything that has a restriction on how namespaces are named other than [this section](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#api:~:text=It%20is%20important%20to%20note,be%20annotated%20with%20the%20label%3A) that talks about namespace run levels.

tylerslaton:
> Looks like there's no kubectl in the container, which causes e2e to fail. Surely we get around this in OLM somehow?

Yeah, we haven't quite finished setting up the E2E suite for this yet IIRC.

manifests/0000_60_cluster-platform-operator-manager_01-rukpak-webhook.certificate.yaml
timflannagan: Yup, works for me.

tylerslaton: Alrighty, those changes are up.

timflannagan: Sounds good :+1:

timflannagan (approved):

timflannagan:
> Looks like there's no kubectl in the container, which causes e2e to fail. Surely we get around this in OLM somehow?

It looks like we already specify `cli: latest` in our ci-operator prowjob configuration: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/platform-operators/openshift-platform-operators-main.yaml#L63. IIRC, that oc binary is just a symlink to kubectl, so we should be able to hotswap any kubectl usage with oc here.

timflannagan:
Looks like no tide.

/close
/re-open

timflannagan:
We're still bootstrapping this repository, so this is a no-op for right now.

/label px-approved
/label qe-approved
/label docs-approved

PR #14, opened 2022-08-19
(psa) label namespace PSA:`restricted`, and make workload compliant

anik120:
This PR applies the `restricted` security level [defined by the Pod Security
Admission Controller](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/). It also modifies
the deployments to add the required securityContext for the pods deployed to be
admitted at the `restricted` level.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -3,6 +3,9 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
+    openshift.io/scc: "anyuid"
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: v1.24
   name: system
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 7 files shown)


config/manager/manager.yaml
```diff
@@ -3,6 +3,9 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
+    openshift.io/scc: "anyuid"
```
tylerslaton, line 6: I'd like to see a comment describing why we're setting these values with a link to the documentation. Do you think that would be too invasive?

config/manager/manager.yaml
```diff
@@ -3,6 +3,9 @@ kind: Namespace
...
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: v1.24
```
tylerslaton, line 8: Same from above applied to these lines.

tylerslaton (approved):
Overall this looks good to me, just a small comment on it. The structure you've gone with here reflects our currently set precedence. We'll need an approve as well as some overrides from @timflannagan when this is ready.

/lgtm

anik120:
/hold 

Till we make necessary changes to rukpak's workloads to make them compatible for the restricted profile.

PR #16, opened 2022-08-22
Designate Platform Operators as Tech Preview

awgreene:
The platform operators project is being released as a Tech Preview
feature within OpenShift. A project managed by CVO is designated as a
Tech Preview feature by setting the "release.openshift.io/feature-gate"
annotation to "TechPreviewNoUpgrade". CVO will not apply any resources
with this annotation and key to the cluster unless the cluster has
Tech Preview features enabled.

Signed-off-by: Alexander Greene <greene.al1991@gmail.com>

### DIFF
--- a/manifests/0000_60_cluster-platform-operator-manager_00-platformoperator.crd.yaml
+++ b/manifests/0000_60_cluster-platform-operator-manager_00-platformoperator.crd.yaml
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.9.0
+    release.openshift.io/feature-gate: TechPreviewNoUpgrade
   creationTimestamp: null
   name: platformoperators.platform.openshift.io
 spec:
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -12,6 +12,10 @@ namePrefix: platform-operators-
 #commonLabels:
 #  someName: someValue
 
+# Annotations to add to all resources.
+commonAnnotations:
+  release.openshift.io/feature-gate: TechPreviewNoUpgrade
+
 bases:
 - ../crd
 - ../rbac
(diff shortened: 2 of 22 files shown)


manifests/0000_60_cluster-platform-operator-manager_00-platformoperator.crd.yaml
```diff
@@ -12,82 +13,82 @@ spec:
     listKind: PlatformOperatorList
     plural: platformoperators
     shortNames:
-    - platform
-    - platforms
+      - platform
+      - platforms
```
awgreene, line 17: Seems like YQ had some opinions on how to format an array, I'll check if I can override this setting.

timflannagan (commented):
LGTM at a glance. I think we may need to double check whether this TechPreviewNoUpgrade value needs to be quoted.

timflannagan:
Example on this field being quoted: https://github.com/openshift/cluster-machine-approver/blob/5237647625ff43d22bf967ea85e5eb54cf9860f8/manifests/01-rbac-capi.yaml#L10. Not too sure if it influences anything.

manifests/0000_60_cluster-platform-operator-manager_00-platformoperator.crd.yaml
awgreene: [Related YQ issue](https://github.com/mikefarah/yq/issues/194)

awgreene:
@timflannagan let me spin up a cluster and see if the feature took affect.

awgreene:
@timflannagan let me spin up a cluster and see if the feature took affect.

manifests/0000_60_cluster-platform-operator-manager_00-platformoperator.crd.yaml
awgreene: Fixed by using [Kustomize commonAnnotations](https://github.com/kubernetes-sigs/kustomize/blob/672c751715be7dd0b43b4a2fce956c84452e0db9/api/types/kustomization.go#L59-L60)

awgreene:
@timflannagan there are [a bunch of instances within the openshift organization](https://github.com/search?q=release.openshift.io%2Ffeature-gate%3A+TechPreviewNoUpgrade&type=code) where the annotation's value does not include the double quote symbole

PR #17, opened 2022-08-22
feat: Proxy Aggregate PO State Back to CVO via a ClusterOperator

exdx:
Implementation of https://hackmd.io/lNm8hwsXR2aAfK29u-riMA


exdx:
See https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md for an example of how to handle the CO object status.

timflannagan (commented):
Can we adding some tests here? I think we'd be able to use unit tests + envtest here to simulate a cluster environment, and be able to test out some basic paths.

cmd/main.go
```diff
@@ -94,6 +96,16 @@ func main() {
 	}
 	//+kubebuilder:scaffold:builder
 
+	// Add Aggregated CO controller to manager
+	if err = (&controllers.AggregatedClusterOperatorReconciler{
+		Client:          mgr.GetClient(),
+		DiscoveryClient: discovery.NewDiscoveryClientForConfigOrDie(ctrl.GetConfigOrDie()),
+		Configv1Client:  configv1client.NewForConfigOrDie(ctrl.GetConfigOrDie()),
```
joelanford, line 103:
```suggestion
		DiscoveryClient: discovery.NewDiscoveryClientForConfigOrDie(mgr.GetConfig()),
		Configv1Client:  configv1client.NewForConfigOrDie(mgr.GetConfig()),
```

internal/aggregated-co/builder.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// WithProgressing sets an OperatorProgressing type condition.
+func (b *Builder) WithProgressing(status configv1.ConditionStatus, message string) *Builder {
+	b.init()
+	condition := &configv1.ClusterOperatorStatusCondition{
+		Type:               configv1.OperatorProgressing,
+		Status:             status,
+		Message:            message,
+		LastTransitionTime: metav1.NewTime(time.Now()),
```
joelanford, line 37: `LastTransitionTime` should only change when the `Status` for this `Type` changes. When we build the status during a reconcile, do we account for the object's existing status conditions to ensure we leave `LastTransitionTime` unmodified when the status doesn't change?

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	}
+	existing, err = w.client.ClusterOperators().Create(context.TODO(), co, metav1.CreateOptions{})
+	return
+}
+
+// UpdateStatus updates the clusteroperator object with the new status specified.
+func (w *Writer) UpdateStatus(existing *configv1.ClusterOperator, newStatus *configv1.ClusterOperatorStatus) error {
+	if newStatus == nil || existing == nil {
+		return errors.New("input specified is <nil>")
```
joelanford, line 57: Should the function signature use concrete values rather than pointers so that we can avoid this `nil` check. 

As an aside, this error seems more like a programmer error than a user error, so perhaps a `panic` is more appropriate than returning it and re-reconciling forever?

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	existingStatus := existing.Status.DeepCopy()
+	if reflect.DeepEqual(existingStatus, newStatus) {
+		return nil
+	}
```
joelanford, line 63: Re: my previous comment about lastTransitionTime, it seems like this will never return `nil` because newStatus will always have different timestamps? (at least the way things are structured now)

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,87 @@
...
+	if _, err := w.client.ClusterOperators().UpdateStatus(context.TODO(), existing, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// IsAPIAvailable return true if cluster operator API is present on the cluster.
+// Otherwise, exists is set to false.
+func (w *Writer) IsAPIAvailable() (exists bool, err error) {
```
joelanford, line 75: Is this function used anywhere?

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,144 @@
...
+			log.Error(err, "error updating CO status")
+		}
+	}()
+
+	// Set the default CO status conditions: Progressing True, Degraded False, Available False
+	builder.WithProgressing(openshiftconfigv1.ConditionTrue, "")
+	builder.WithDegraded(openshiftconfigv1.ConditionFalse)
+	builder.WithAvailable(openshiftconfigv1.ConditionFalse, "", "")
+
+	POList := &platformv1alpha1.PlatformOperatorList{}
```
joelanford, line 79: nit:  don't capitalize non-global variable names.
```suggestion
	poList := &platformv1alpha1.PlatformOperatorList{}
```

joelanford (commented):
Looks like this is really shaping up! Nice!

Pet peave (and I realize completely out of your control): I really dislike `vendor` directories in downstream code. I understand the reasoning, but its super annoying that it shows up in PRs with things like "226 files changed"... smh

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	// TODO: Setup predicate filter to only watch the Aggregate CO and ignore other COs
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&openshiftconfigv1.ClusterOperator{}).
+		Watches(&source.Kind{Type: &platformv1alpha1.PlatformOperator{}}, handler.EnqueueRequestsFromMapFunc(util.RequeueBundleDeployment(mgr.GetClient()))).
+		Complete(a)
+}
+
+type POStatusErrors struct {
+	FailingPOs    []*platformv1alpha1.PlatformOperator
```
tylerslaton, line 116: It does not seem that we are using this property on `POStatusErrors` in the code anywhere. Is there value in storing it to memory that I'm not seeing? Is this use preparing for a use case in the future?

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,144 @@
...
+
+type AggregatedClusterOperatorReconciler struct {
+	client.Client
+	DiscoveryClient discovery.DiscoveryInterface
+	Configv1Client  configv1client.ConfigV1Interface
+}
+
+const aggregateCOName = "platform-operators-aggregated"
+
+// TODO kubebuilder markers
```
tylerslaton, line 48: Should we address this in this PR?

internal/aggregated-co/builder.go
exdx: Yes, we should definitely do that

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := logr.FromContext(ctx)
+	log.Info("reconciling request", "req", req.NamespacedName)
+	defer log.Info("finished reconciling request", "req", req.NamespacedName)
+
+	if req.Name != aggregateCOName {
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
exdx: I'd like to do that yes. I couldn't quite figure it out yesterday but I can try again. If you have an example where the watch references an object by name that would be very helpful.

internal/aggregated-co/writer.go
exdx: This came from the OLM project, I can update it

controllers/aggregated_clusteroperator_controller.go
exdx: Potentially, yes. I can slim down the data structure but I imagine in the future we would have more robust status messaging

tylerslaton:
Talked about this offline but wanted to make sure we're tracking that we need to add to the new CO manifest to the `config` directory so that `kustomize` generates it and then copy that to the `manifests` dir. This should fix this `verify` failure.

Makefile
```diff
@@ -94,6 +94,7 @@ manifests: generate kustomize
 	$(MV_TMP_DIR)/v1_serviceaccount_platform-operators-controller-manager.yaml manifests/0000_60_cluster-platform-operator-manager_01-serviceaccount.yaml
 	$(MV_TMP_DIR)/v1_service_platform-operators-controller-manager-metrics-service.yaml manifests/0000_60_cluster-platform-operator-manager_02-metricsservice.yaml
 	$(MV_TMP_DIR)/apps_v1_deployment_platform-operators-controller-manager.yaml manifests/0000_60_cluster-platform-operator-manager_06-deployment.yaml
+	$(MV_TMP_DIR)/config.openshift.io_v1_clusteroperator_platform-operators-aggregated.yaml manifests/0000_60_cluster-platform-operator-manager_07-aggregated-clusteroperator.yaml
```
tylerslaton, line 97: 🎉

exdx:
> Talked about this offline but wanted to make sure we're tracking that we need to add to the new CO manifest to the `config` directory so that `kustomize` generates it and then copy that to the `manifests` dir. This should fix this `verify` failure.

Yes, it should be resolved in the latest force push, see [a7cdb4e](https://github.com/openshift/platform-operators/pull/17/commits/a7cdb4ecbd46b4f00b9b9668d04b4ac7052837dd)

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (a *AggregatedClusterOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// TODO: Setup predicate filter to only watch the Aggregate CO and ignore other COs
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&openshiftconfigv1.ClusterOperator{}).
```
timflannagan, line 110: I think we'd could do something like the following:

```go
		For(&openshiftconfigv1.ClusterOperator{}, builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool {
			return object.GetName() == aggregateCOName
		}))).
```

tylerslaton:
After the most recent commits this looks good to me.

/lgtm

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Writer encapsulates logic for cluster operator object API. It is used to
+// update ClusterOperator resource.
+type Writer struct {
+	client configv1client.ConfigV1Interface
+}
+
+// EnsureExists ensures that the cluster operator resource exists with a default
+// status that reflects expecting status.
+func (w *Writer) EnsureExists(name string) (existing *configv1.ClusterOperator, err error) {
```
joelanford, line 30: Always pass a context in from the function signature if you ever need to call anything that accepts a context.
```suggestion
func (w *Writer) EnsureExists(ctx context.Context, name string) (existing *configv1.ClusterOperator, err error) {
```

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,66 @@
...
+		panic("input specified is <nil>")
+	}
+
+	existingStatus := existing.Status.DeepCopy()
+	if reflect.DeepEqual(existingStatus, newStatus) {
+		return nil
+	}
+
+	existing.Status = *newStatus
+	if _, err := w.client.ClusterOperators().UpdateStatus(context.TODO(), existing, metav1.UpdateOptions{}); err != nil {
```
joelanford, line 61: Same comment as above: pass a context in and use it.

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Writer encapsulates logic for cluster operator object API. It is used to
+// update ClusterOperator resource.
+type Writer struct {
+	client configv1client.ConfigV1Interface
+}
+
+// EnsureExists ensures that the cluster operator resource exists with a default
+// status that reflects expecting status.
```
joelanford, line 29: Looks like this function is also unused? (We're going to include the CO in our payload manifests, so I think it's safe to assume that it will exist)

joelanford (commented):
Did the "don't touch the last transition time for conditions that didn't change" problem get addressed yet?

Maybe this is a good area on which to focus tests for the builder/writer.

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	client.Client
+	DiscoveryClient discovery.DiscoveryInterface
+	Configv1Client  configv1client.ConfigV1Interface
+}
+
+const aggregateCOName = "platform-operators-aggregated"
+
+//+kubebuilder:rbac:groups=platform.openshift.io,resources=platformoperators,verbs=get;list;watch
+//+kubebuilder:rbac:groups=platform.openshift.io,resources=platformoperators/status,verbs=get
+//+kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators,verbs=get;list;watch;create;update;patch;delete
```
timflannagan, line 51: nit: Do we need access to all these verbs, or can we reduce them to the subset we need for this controller?

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,148 @@
...
+// A nil return value indicates no errors were found with the POs provided.
+func (a *AggregatedClusterOperatorReconciler) inspectPlatformOperators(POList *platformv1alpha1.PlatformOperatorList) *POStatusErrors {
+	POstatuses := new(POStatusErrors)
+
+	for _, po := range POList.Items {
+		po := po.DeepCopy()
+		status := po.Status
+
+		for _, condition := range status.Conditions {
+			if condition.Reason == platformv1alpha1.ReasonSourceFailed || condition.Reason == platformv1alpha1.ReasonApplyFailed {
```
timflannagan, line 135: nit: I think we'll end up consolidating this into a single status condition. I'd update this to just care about the ReasonApplyFailed for now. Alternatively, we could just iterate over the list of POs returned from the list query, and check whether `meta.FindStatusCondition(po.Status.Conditions, platformv1alpha1.TypeApplied`) is reporting a non metav1.ConditionTrue status.

internal/aggregated-co/writer.go
```diff
@@ -0,0 +1,66 @@
...
+func NewWriter(client configv1client.ConfigV1Interface) *Writer {
+	return &Writer{
+		client: client,
+	}
+}
+
+// Writer encapsulates logic for cluster operator object API. It is used to
+// update ClusterOperator resource.
+type Writer struct {
+	client configv1client.ConfigV1Interface
```
timflannagan, line 25: nit: We could refactor this to use c-r's dynamic client, and add the configv1 scheme to a central scheme?

internal/aggregated-co/writer.go
timflannagan: I think we've seen a couple of CO controller implementations (including OLM/marketplace) have logic for creating the resource during apierrors.IsNotFound(err) error checks. Not too sure whether that was a legacy issue, and that kind of logic is no longer needed, or it's something that the CVO team recommends as a best practice.

In any case, this kind of logic seems like it could be refactored to use c-r's CreateOrUpdate logic?

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -0,0 +1,148 @@
...
+
+	poList := &platformv1alpha1.PlatformOperatorList{}
+	if err := a.List(ctx, poList); err != nil {
+		log.Error(err, "error listing platformoperators")
+		return ctrl.Result{}, err
+	}
+
+	if len(poList.Items) == 0 {
+		// No POs on cluster, everything is fine
+		coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "", "No POs Found")
```
timflannagan, line 91: nit: It looks like WithAvailable is expecting a status condition reason here which are typically in the form of camel-cased, one-word phrases.

controllers/aggregated_clusteroperator_controller.go
kuiwang02, line 135: @exdx if the platform operator is being installed, there is no result yet.
maybe 
1, there is no condition yet in the PO .status.conditions.
or
2, there is condition in the PO .status.conditions, but the reason is not Failed.

For these situations, the current code treat it as "All POs in a successful state". 
So it is negative PO status checking.

please correct me if my understanding is wrong. thanks.

internal/aggregated-co/writer.go
timflannagan: This about this more: I think this is less of a nit and more of a potential blocker here. We're introducing a dependency that we don't necessarily need, and inflating the changeset of this PR given we're committing the vendor tree in this repository.

internal/aggregated-co/writer.go
exdx: I think the context here is that in OLM, this code was meant to safely run upstream, where no CO would be present. It's not being used by this PR, so I'm simply going to remove this function and keep things simple.

exdx: Since the aggregate CO is part of the PO payload, we can rely on CVO to ensure it's always on-cluster.

internal/aggregated-co/writer.go
exdx: I'm not sure I see the benefit of switching to the dynamic client here -- having the typed client in tree will be helpful, especially as we will have to manage the POM CO at some point too. We do vendor a bit, but just the typed client we need essentially.

controllers/aggregated_clusteroperator_controller.go
exdx: By default, when the controller enters the loop, it sets the following status:
```
	coBuilder.WithProgressing(openshiftconfigv1.ConditionTrue, "")
	coBuilder.WithDegraded(openshiftconfigv1.ConditionFalse)
	coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, "", "")
``

If a PO is progressing, and not in an error state, the status will be updated to
```
	coBuilder.WithProgressing(openshiftconfigv1.ConditionTrue, "")
	coBuilder.WithDegraded(openshiftconfigv1.ConditionFalse)
	coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "All POs in a successful state", "POsHealthy")
```

So yes, the Available/Progressing condition will bet set to true. The code is just checking for failures essentially.

Comment (exdx) [reply controllers/aggregated_clusteroperator_controller.go]: We could also check if all the POs are in a successful terminal state, and only then update Available=True? Does that make sense @joelanford @timflannagan

**Comment by exdx** [comment]
> Did the "don't touch the last transition time for conditions that didn't change" problem get addressed yet?
> 
> Maybe this is a good area on which to focus tests for the builder/writer.

Yes, I added some logic to the condition handlers to check if the status is matching

**Review by tylerslaton — COMMENTED** [review]
Took another look over this and everything looks good to me.

/lgtm

File: controllers/aggregated_clusteroperator_controller.go [hunk 19]
```diff
@@ -100,6 +101,6 @@ func (a *AggregatedClusterOperatorReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&openshiftconfigv1.ClusterOperator{}, builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool {
 			return object.GetName() == aggregateCOName
 		}))).
-		Watches(&source.Kind{Type: &platformv1alpha1.PlatformOperator{}}, handler.EnqueueRequestsFromMapFunc(util.RequeueBundleDeployment(mgr.GetClient()))).
```
timflannagan: This was needed as I saw a panic in the aggregate CO controller when deploying these changes to a crc cluster. Updated the logic to instead requeue the aggregate CO resource during PO watch events.

internal/sourcer/registry_sourcer.go
```diff
@@ -54,7 +54,7 @@ func (cs catalogSource) Source(ctx context.Context, po *platformv1alpha1.Platfor
 }
 
 func (s sources) GetCandidates(ctx context.Context, po *platformv1alpha1.PlatformOperator) (bundles, error) {
-	if len(s) > 1 {
+	if len(s) != 1 {
```
timflannagan, line 57: I saw a panic when there were no available catalog sources in the cluster. This is a follow-up to the changes made in #26.

timflannagan:
Adding the requisite labels as this is a pre-requisite to introducing the PO components to the payload (#23), which will follow the no-ff process.

/label px-approved
/label docs-approved
/label qe-approved

tylerslaton (commented):
/lgtm

timflannagan (approved):

PR #23, opened 2022-08-31
Add the CPOM and rukpak components to the payload

timflannagan:
/wip

PR #28, opened 2022-09-02
feat: introduce the PSA changes from upstream RukPak

tylerslaton:
Bringing the upstream RukPak manifest changes downstream to try and unblock #23

### DIFF
--- a/config/rukpak/apis/webhooks/resources/deployment.yaml
+++ b/config/rukpak/apis/webhooks/resources/deployment.yaml
@@ -13,9 +13,17 @@ spec:
       labels:
         app: webhooks
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: webhooks-admin
       containers:
         - name: webhooks
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: [ "ALL" ]
           command: ["/webhooks"]
           image: quay.io/operator-framework/rukpak:main
           imagePullPolicy: IfNotPresent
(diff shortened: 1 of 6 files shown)


config/rukpak/apis/webhooks/resources/deployment.yaml
```diff
@@ -13,9 +13,17 @@ spec:
       labels:
         app: webhooks
     spec:
+      securityContext:
```
timflannagan, line 16: Is spec.securityContext a valid DeploymentSpec field:

```bash
$ k explain deploy.spec | grep securityContext
```

Returns nothing. But deploy.spec.template.spec does appear to have a securityContext field.

tylerslaton: This `securityContext` is already defined inside of the `deploy.spec.template.spec` but it is hard to see because of how GitHub is rendering it. You can hit the little arrow button above this to see.

timflannagan: Yeah my bad. What's even worse is that I pressed the drop down, and still mixed this up.

/approve
/lgtm

timflannagan:
It would also be nice if we could figure out a way to use kustomize patches for this. But that can be done in a follow-up anyways.

tylerslaton:
> It would also be nice if we could figure out a way to use kustomize patches for this. But that can be done in a follow-up anyways.

That's a good call, I'll make an issue for it. Shouldn't be too hard.

timflannagan:
This is blocking the #23 PR which introduces the rukpak/CPOM components to the payload. Adding the required labels in the meantime.

/label px-approved
/label docs-approved
/label qe-approved

exdx (approved):

PR #31, opened 2022-09-02
Makefile,manifests: Add hack for injecting the self managed HA cluster cluster

timflannagan:
Update the PlatformOperator CRD YAML manifests housed in o/api, and inject the required self-managed HA cluster profile annotation. This is a short term hack in order to unblock payload introduction.


tylerslaton (approved):
Should definitely make sure we follow up with a quick fix once that PR merges.
/lgtm

PR #34, opened 2022-09-06
controllers,internal: Refactor the logic for inspecting POs

timflannagan:

### DIFF
--- a/controllers/aggregated_clusteroperator_controller.go
+++ b/controllers/aggregated_clusteroperator_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 
 	openshiftconfigv1 "github.com/openshift/api/config/v1"
-	utilerror "k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -87,7 +86,7 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
 	if statusErrorCheck := util.InspectPlatformOperators(poList); statusErrorCheck != nil {
-		coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, utilerror.NewAggregate(statusErrorCheck.FailingErrors).Error(), "POError")
+		coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, statusErrorCheck.Error(), "POError")
 		return ctrl.Result{}, nil
 	}
 	coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "All POs in a successful state", "POsHealthy")
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -7,7 +7,10 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilerror "k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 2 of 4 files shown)


controllers/aggregated_clusteroperator_controller.go
```diff
@@ -87,8 +86,8 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
 	if statusErrorCheck := util.InspectPlatformOperators(poList); statusErrorCheck != nil {
-		coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, utilerror.NewAggregate(statusErrorCheck.FailingErrors).Error(), "POError")
-		return ctrl.Result{}, nil
+		coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, statusErrorCheck.Error(), "POError")
+		return ctrl.Result{}, statusErrorCheck
```
timflannagan, line 90: Forgot to revert this change. I realized this was causing hotloop-esq behaviors, and we already established the requisite watches to get any new information.

internal/util/util.go
```diff
@@ -74,33 +77,37 @@ func RequeueClusterOperator(c client.Client, name string) handler.MapFunc {
...
 
+// inspectPlatformOperator is responsible for inspecting an individual platform
+// operator resource, and determining whether it's reporting any failing conditions.
+// In the case that the PO resource is expressing failing states, then an error
+// will be returned to reflect that.
+func inspectPlatformOperator(po platformv1alpha1.PlatformOperator) error {
+	applied := meta.FindStatusCondition(po.Status.Conditions, platformtypes.TypeApplied)
+	if applied == nil {
+		return buildPOFailureMessage(po.GetName(), platformtypes.ReasonApplyPending)
+	}
```
timflannagan, line 104: I _think_ this is the correct behavior but I'm open to other's thoughts here. Basically if there's no TypeApplied status condition type on the PO resource we're inspecting, then we can make the assumption that either the CPOM controller hasn't reconciled it yet, or there's a core issue in the CPOM controller that prevents that resource from being inspected.

exdx (approved):
/lgtm

exdx:
Thanks for cleaning this up a bit

timflannagan:
Adding the required labels as this is a prerequisite to #23 which will follow the no-ff process.

/label qe-approved
/label px-approved
/label docs-approved

internal/util/util.go
```diff
@@ -74,33 +77,37 @@ func RequeueClusterOperator(c client.Client, name string) handler.MapFunc {
...
+	if applied.Status != metav1.ConditionTrue {
+		return buildPOFailureMessage(po.GetName(), applied.Reason)
+	}
 	return nil
 }
+
+func buildPOFailureMessage(name, reason string) error {
+	return fmt.Errorf("encountered the failing %s platform operator with reason %q", name, reason)
+}
```
tylerslaton, line 113: Should we consider a PO to be failing if it is pending?

internal/util/util_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+	platformtypes "github.com/openshift/platform-operators/api/v1alpha1"
+)
+
+func Test_inspectPlatformOperator(t *testing.T) {
```
tylerslaton, line 12: Is this name formatted this way for a reason? Why not just `TestInspectPlatformOperator`?

internal/util/util_test.go
```diff
@@ -0,0 +1,108 @@
...
+						},
+					},
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := inspectPlatformOperator(tt.args.po); (err != nil) != tt.wantErr {
```
tylerslaton, line 103: Should we check what the error is and confirm what we're expecting? I mainly ask since we return a different error for a failing PO and for a pending one.

tylerslaton (approved):
Nothing blocking, just a few thoughts.

/lgtm

tylerslaton:
So you can take a look at the comments.

/hold

internal/util/util.go
tylerslaton: Mainly just asking for the message here/

internal/util/util_test.go
timflannagan: This was auto-generated for me by vscode. We can refactor down the line if needed.

internal/util/util_test.go
timflannagan: Yeah that sounds like the right test check. Just trying to get something in the door so we have a baseline while iterating. I can revisit fleshing this test case out in the future.

timflannagan:
Comments are valid but we're still iterating and be tackled in a follow-up.

/hold cancel

PR #35, opened 2022-09-06
tests: Add aggregated clusteroperator e2e test

exdx:
Adds e2e tests for the aggregated clusteroperator controller.


timflannagan (commented):
Can we ditch the ClusterOperator CRD from these changes? It should remove the apiextensions-apiserver direct dependency, and we avoid risking any skew between that CRD and what's present on the cluster by default.

If we're trying to get all of this working on a kind cluster, can we handle this instead through a bash script and document this approach for any dev workflows?

timflannagan:
Also: register the configv1 scheme in e2e_suite_test.go?

test/e2e/aggregated_clusteroperator_test.go
```diff
@@ -0,0 +1,145 @@
...
+				}).Should(And(
+					Not(BeNil()),
+					WithTransform(func(c *configv1.ClusterOperatorStatusCondition) configv1.ClusterStatusConditionType { return c.Type }, Equal(configv1.OperatorAvailable)),
+					WithTransform(func(c *configv1.ClusterOperatorStatusCondition) configv1.ConditionStatus { return c.Status }, Equal(configv1.ConditionTrue)),
+					WithTransform(func(c *configv1.ClusterOperatorStatusCondition) string { return c.Reason }, Equal("POsHealthy")),
+					WithTransform(func(c *configv1.ClusterOperatorStatusCondition) string { return c.Message }, ContainSubstring("All POs in a successful state")),
+				))
+
+			})
+		})
```
tylerslaton, line 143: Can we add two test cases here?

1. One PO that fails reports an unhealthy CO to the CVO
2. One PO fails out of multiple succeeding POs which reports an unhealthy CO to CVO

timflannagan: I think we'd want to initially test out the following scenarios:

- Happy path
- No POs in the cluster
- A PO that's failing to find a bundle from the desired package name
- A mixture of successful and failed POs results in an unavailable condition

These are likely closer to unit tests, but we can work towards refactoring those checks further down the line.

timflannagan:
Tested these changes manually on a crc cluster. I ran `make test-e2e TEST="aggregated clusteroperator controller"` to isolate these tests given the POC e2e tests still need to be updated.

timflannagan:
These changes only affect our testing environment vs. anything that's user facing, so I'm going to preemptively apply the requisite labels.

/label px-approved
/label qe-approved
/label docs-approved

test/e2e/aggregated_clusteroperator_test.go
timflannagan: Done. Added all of those test cases in the latest force push.

timflannagan:
I rebuild the container image locally, and re-ran the testing suite to verify the changes. This seems sufficient to me for now.

/lgtm

PR #40, opened 2022-09-07
minor cleanup and refactoring for consistency and simplicity

joelanford:
Somewhat scatter-brained refactor/fixup PR for random things I found. I'll clean this up more if desired.

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

### DIFF
--- a/controllers/aggregated_clusteroperator_controller.go
+++ b/controllers/aggregated_clusteroperator_controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 
 	configv1 "github.com/openshift/api/config/v1"
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -62,19 +64,26 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	}
 	defer func() {
 		if err := coWriter.UpdateStatus(ctx, aggregatedCO, coBuilder.GetStatus()); err != nil {
-			log.Error(err, "error updating CO status")
+			log.Error(err, "error updating cluster operator status")
 		}
 	}()
 
 	// Set the default CO status conditions: Progressing=True, Degraded=False, Available=False
-	coBuilder.WithProgressing(configv1.ConditionTrue, "")
-	coBuilder.WithDegraded(configv1.ConditionFalse)
-	coBuilder.WithAvailable(configv1.ConditionFalse, "", "")
+	// TODO: always set a reason (message is optional, but desirable)
+	coBuilder.WithProgressing(metav1.ConditionTrue, "", "")
+	coBuilder.WithDegraded(metav1.ConditionFalse, "", "")
+	coBuilder.WithAvailable(metav1.ConditionFalse, "", "")
 	coBuilder.WithVersion("operator", r.ReleaseVersion)
-	coBuilder.WithRelatedObject("", "namespaces", "", r.SystemNamespace)
-	coBuilder.WithRelatedObject("platform.openshift.io", "platformoperators", "", "")
-	coBuilder.WithRelatedObject("core.rukpak.io", "bundles", "", "")
-	coBuilder.WithRelatedObject("core.rukpak.io", "bundledeployments", "", "")
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: "", Resource: "namespaces", Name: r.SystemNamespace})
+
+	// NOTE: Group and resource can be referenced without name/namespace set, which is a signal
+	// that _ALL_ objects of that group/resource are related objects. This is useful for
+	// must-gather automation.
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: platformv1alpha1.GroupName, Resource: "platformoperators"})
+
+	// TODO: move platform operator ownership of rukpak objects out prior to rukpak or PO GA.
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundles"})
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundledeployments"})
 
 	poList := &platformv1alpha1.PlatformOperatorList{}
 	if err := r.List(ctx, poList); err != nil {
@@ -82,8 +91,12 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	}
 	if len(poList.Items) == 0 {
 		// No POs on cluster, everything is fine
-		coBuilder.WithAvailable(configv1.ConditionTrue, "No POs are present in the cluster", "NoPOsFound")
-		coBuilder.WithProgressing(configv1.ConditionFalse, "No POs are present in the cluster")
+		// TODO: cleanup condition reasons.
+		//   1. use constants for condition reasons.
+		//   2. discuss/agree on reason values.
+		//   3. don't use abbreviations.
+		coBuilder.WithAvailable(metav1.ConditionTrue, "NoPOsFound", "No POs are present in the cluster")
+		coBuilder.WithProgressing(metav1.ConditionFalse, "", "No POs are present in the cluster")
 		return ctrl.Result{}, nil
 	}
 
@@ -91,11 +104,11 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
 	if statusErrorCheck := util.InspectPlatformOperators(poList); statusErrorCheck != nil {
-		coBuilder.WithAvailable(configv1.ConditionFalse, statusErrorCheck.Error(), "POError")
+		coBuilder.WithAvailable(metav1.ConditionFalse, "POError", statusErrorCheck.Error())
 		return ctrl.Result{}, nil
 	}
-	coBuilder.WithAvailable(configv1.ConditionTrue, "All POs in a successful state", "POsHealthy")
-	coBuilder.WithProgressing(configv1.ConditionFalse, "All POs in a successful state")
+	coBuilder.WithAvailable(metav1.ConditionTrue, "POsHealthy", "All POs in a successful state")
+	coBuilder.WithProgressing(metav1.ConditionFalse, "", "All POs in a successful state")
 
 	return ctrl.Result{}, nil
 }
--- a/controllers/aggregated_clusteroperator_controller.go
+++ b/controllers/aggregated_clusteroperator_controller.go
@@ -74,16 +74,9 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	coBuilder.WithDegraded(metav1.ConditionFalse, "", "")
 	coBuilder.WithAvailable(metav1.ConditionFalse, "", "")
 	coBuilder.WithVersion("operator", r.ReleaseVersion)
-	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: "", Resource: "namespaces", Name: r.SystemNamespace})
 
-	// NOTE: Group and resource can be referenced without name/namespace set, which is a signal
-	// that _ALL_ objects of that group/resource are related objects. This is useful for
-	// must-gather automation.
-	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: platformv1alpha1.GroupName, Resource: "platformoperators"})
-
-	// TODO: move platform operator ownership of rukpak objects out prior to rukpak or PO GA.
-	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundles"})
-	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundledeployments"})
+	// Set the static set of related objects
+	setStaticRelatedObjects(coBuilder, r.SystemNamespace)
 
 	poList := &platformv1alpha1.PlatformOperatorList{}
 	if err := r.List(ctx, poList); err != nil {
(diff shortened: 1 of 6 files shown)


timflannagan:
Looks reasonable to me at a glance. This will fail the linting check due to the package import grouping, and I added some unit tests for when the TypeApplied condition is nil, so you'd have to update those as well.

internal/util/util.go
```diff
@@ -77,27 +78,27 @@ func RequeueClusterOperator(c client.Client, name string) handler.MapFunc {
...
-	return nil
+	return utilerror.NewAggregate(poErrors)
 }
 
 // inspectPlatformOperator is responsible for inspecting an individual platform
 // operator resource, and determining whether it's reporting any failing conditions.
 // In the case that the PO resource is expressing failing states, then an error
 // will be returned to reflect that.
 func inspectPlatformOperator(po platformv1alpha1.PlatformOperator) error {
+	if equality.Semantic.DeepEqual(po.Status, platformv1alpha1.PlatformOperatorStatus{}) {
```
kuiwang02, line 99: @joelanford Hi, when `equality.Semantic.DeepEqual(po.Status, platformv1alpha1.PlatformOperatorStatus{})` return True, does it mean the PO status part is empty? 
if it is correct understanding, I can not understand why `return nil` here because I think`return nil` means PO is in healthy status in method `InspectPlatformOperators `, and not report error to aggregated CO. But we do not the PO final status yet (maybe finally the PO is not installed successfully). So, here maybe `return buildPOFailureMessage(po.GetName(), "PO has no status field")`

(I just worry about that before PO installation is done (no final status yet), we report wrong information to aggregated CO. Thanks to check my concern)

joelanford: > does it mean the PO status part is empty?

Yes

> I can not understand why return nil here 

@timflannagan and I were chatting about this yesterday. In general, a PO with no status is a PO that has yet to be reconciled by the PO controller and that typically happens only as the PO is _just_ created. When the PO controller reconciles this PO, it'll attach a status and then this controller will reconcile again.

One thought I had to guard against the possibility of the PO controller never actually reconciling the PO: We could check `status is empty` and `creationTimestamp is within the last 1 minute`.

Which would result in:
1. Empty status within a minute of being created: no problem, return `nil`
2. Empty status after a minute of being create: something's wrong, return an error.

WDYT?

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,34 +63,40 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
...
 	if err := a.List(ctx, poList); err != nil {
 		return ctrl.Result{}, err
 	}
 	if len(poList.Items) == 0 {
 		// No POs on cluster, everything is fine
-		coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "No POs are present in the cluster", "NoPOsFound")
-		coBuilder.WithProgressing(openshiftconfigv1.ConditionTrue, "No POs are present in the cluster")
+
+		// TODO: cleanup condition reasons.
+		//   1. use constants for condition reasons.
```
tylerslaton, line 84: We have some condition constants already defined but it doesn't seem like we are using them here for some reason. Maybe we just need to add onto that list?

https://github.com/openshift/platform-operators/blob/03149b26e54fa0737b37a9e183bb75fbf73ea94a/api/v1alpha1/types.go#L22-L25

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,34 +63,40 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
...
+		coBuilder.WithProgressing(metav1.ConditionTrue, "", "No POs are present in the cluster")
 		return ctrl.Result{}, nil
 	}
 
 	// check whether any of the underlying PO resources are reporting
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
 	if statusErrorCheck := util.InspectPlatformOperators(poList); statusErrorCheck != nil {
-		coBuilder.WithAvailable(openshiftconfigv1.ConditionFalse, statusErrorCheck.Error(), "POError")
+		coBuilder.WithAvailable(metav1.ConditionFalse, "POError", statusErrorCheck.Error())
```
tylerslaton, line 96: In general should we be using the `apimachinery` api's for this instead of the `OpenShift` ones? Mainly asking to see how you decided to change this.

tylerslaton (approved):
Nice work!

/lgtm

internal/util/util.go
timflannagan: @joelanford Should we add a comment that basically explains what you had written down here? I should've done that in the PR that refactored these checks.

In general, I think that option you outlined seems reasonable at a glance but introduces potential complexity.

internal/util/util.go
```diff
@@ -77,27 +78,27 @@ func RequeueClusterOperator(c client.Client, name string) handler.MapFunc {
...
 		if err := inspectPlatformOperator(po); err != nil {
-			failingPOs = append(failingPOs, err)
+			poErrors = append(poErrors, err)
 		}
 	}
-	if len(failingPOs) > 0 {
-		return utilerror.NewAggregate(failingPOs)
-	}
-	return nil
+	return utilerror.NewAggregate(poErrors)
```
timflannagan, line 91: :+1: Whoops. This was a bit obvious in hindsight :sweat_smile:.

timflannagan:
lgtm

internal/util/util.go
timflannagan: Thinking about this more, in the current implementation I think we'll see flapping in the aggregate ClusterOperator resource we're managing. I'm working on the POM component's implementation, which is responsible for bubbling up the underlying BundleDeployment status. In the case where the BD resource is waiting to be unpacked, we'll effectively see there's no TypeApplied status condition type present yet, and propagate the ApplyPending status condition reason in the PO resource. The aggregate CO controller would see that, and mark the aggregate CO resource with the following:

```bash
$ k get co -w
...
platform-operators-aggregated                        False       True          False      0s      encountered the failing cincinnati-operatorrwxr4 platform operator with reason "ApplyPending"
...
platform-operators-aggregated                        False       True          False      0s      encountered the failing cincinnati-operatorrwxr4 platform operator with reason "ApplyPending"
platform-operators-aggregated                        True        True          False      0s      
```

Which feels like the wrong behavior, UX, etc.

controllers/aggregated_clusteroperator_controller.go
joelanford: Those are reasons for the PO `Applied` condition type. These are reasons for the ClusterOperator condition types.

controllers/aggregated_clusteroperator_controller.go
joelanford: They're both definitions of basically the exact same thing. I changed this mainly because apimachinery has some nice helpers for interacting with conditions that helps us avoid maintaining complexity around the "rules" of conditions, like knowing when to bump last transition time.

internal/util/util.go
kuiwang02: @joelanford Thanks to check my comment.

Your idea is ok, but potential complexity (same to timflannagan).
maybe we just change `return nil` to `return buildPOFailureMessage(po.GetName(), platformtypes.ReasonApplyPending)` because we possible treat `no reconciled yet` as `ApplyPending`.

anyway I am also OK if you want to take your idea on 1minute.

joelanford: My main concern is that it seems misleading and alarmist to say a PO is failing when in fact it has just been created and is waiting for the PO controller to do something with it.

In the most recent commit, I've updated to include the "time since creationTimestamp" check. But if there's another approach that lets us avoid that complexity AND avoid saying the PO is failing, I'm totally willing to get rid of that.

internal/util/util.go
```diff
@@ -71,27 +73,28 @@ func RequeueClusterOperator(c client.Client, name string) handler.MapFunc {
...
 }
 
 // inspectPlatformOperator is responsible for inspecting an individual platform
 // operator resource, and determining whether it's reporting any failing conditions.
 // In the case that the PO resource is expressing failing states, then an error
 // will be returned to reflect that.
 func inspectPlatformOperator(po platformv1alpha1.PlatformOperator) error {
+	if equality.Semantic.DeepEqual(po.Status, platformv1alpha1.PlatformOperatorStatus{}) &&
+		(po.CreationTimestamp.IsZero() || time.Since(po.CreationTimestamp.Time) < time.Minute) {
```
joelanford, line 95: Just want to highlight that my rebase included this addition based on the earlier discussion here: https://github.com/openshift/platform-operators/pull/40#discussion_r964424982

timflannagan:
The e2e results are a no-op until #58 merges.

timflannagan:
Now that the payload PR has merged, does it make sense to track this as a bug somewhere? I'm not sure we have a decent Jira home right now, so maybe an upstream issue could be sufficient here.

timflannagan:
/retitle OLM-2776 minor cleanup and refactoring for consistency and simplicity

timflannagan:
/retitle OCPBUGS-1477 minor cleanup and refactoring for consistency and simplicity

joelanford:
/hold

I need to think through this CVO doc line item before this merges:

> An operator should not report the Available status condition the first time until they are completely rolled out (or within some reasonable percentage if the component must be installed to all nodes)

If we consider just-created, but not-yet-reconciled POs to be Available, we'll signal to CVO that the POs have been "completely rolled out" prior to the rollout even starting. It seems like we may need some logic that:
1. During a cluster install scenario, ensures all POs have BDs that have progressed to `InstallationSucceeded` before setting A=T
2. After initial cluster install, any newly created POs should trigger `Progressing=true` until that newly created PO's BD has attempted an install.

joelanford:
I reverted the changes to the way the Available condition is calculated. For now, that means we'll keep the existing logic which means that a PO that is missing the `Installed` condition will be considered Available=False.

joelanford:
I also disassociated this PR from OCPBUGS-1477 since it no longer addresses that bug. And I actually don't think that's a bug in the first place. So I think I'd suggest that we capture any takeaways and close it as "not a bug"

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,40 +64,51 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
...
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
 	if statusErrorCheck := util.InspectPlatformOperators(poList); statusErrorCheck != nil {
-		coBuilder.WithAvailable(configv1.ConditionFalse, statusErrorCheck.Error(), "POError")
+		coBuilder.WithAvailable(metav1.ConditionFalse, "POError", statusErrorCheck.Error())
 		return ctrl.Result{}, nil
 	}
-	coBuilder.WithAvailable(configv1.ConditionTrue, "All POs in a successful state", "POsHealthy")
-	coBuilder.WithProgressing(configv1.ConditionFalse, "All POs in a successful state")
+	coBuilder.WithAvailable(metav1.ConditionTrue, "POsHealthy", "All POs in a successful state")
```
timflannagan, line 110: I've seen a bunch of other core ClusterOperator resources set the "AsExpected" reason during success states.

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,40 +64,51 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
...
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: "", Resource: "namespaces", Name: r.SystemNamespace})
+
+	// NOTE: Group and resource can be referenced without name/namespace set, which is a signal
+	// that _ALL_ objects of that group/resource are related objects. This is useful for
+	// must-gather automation.
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: platformv1alpha1.GroupName, Resource: "platformoperators"})
+
+	// TODO: move platform operator ownership of rukpak objects out prior to rukpak or PO GA.
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundles"})
+	coBuilder.WithRelatedObject(configv1.ObjectReference{Group: rukpakv1alpha1.GroupVersion.Group, Resource: "bundledeployments"})
```
timflannagan, line 86: We could also move this logic to a method/function that adds these related objects to the builder instance.

timflannagan:
The refactoring generally lgtm. It would be nice to add constants for the common set of reasons that we expect to use here.

joelanford:
Latest commit covers your latest round of feedback @timflannagan.

joelanford:
> Another potential TODO is updating the aggregate CO implementation to re-create that resource when it no longer exists.

Based on what Trevor said yesterday, I kinda think we leave it up to CVO to recreate the CO.

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,40 +64,43 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	}
 	defer func() {
 		if err := coWriter.UpdateStatus(ctx, aggregatedCO, coBuilder.GetStatus()); err != nil {
-			log.Error(err, "error updating CO status")
+			log.Error(err, "error updating cluster operator status")
```
tylerslaton, line 67: nit:
```suggestion
			log.Error(err, "error updating ClusterOperator status")
```

All other instances apply if we want to change this.

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -62,40 +64,43 @@ func (r *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
...
+		coBuilder.WithProgressing(metav1.ConditionFalse, clusteroperator.ReasonAsExpected, "No platform operators are present in the cluster")
 		return ctrl.Result{}, nil
 	}
 
 	// check whether any of the underlying PO resources are reporting
 	// any failing status states, and update the aggregate CO resource
 	// to reflect those failing PO resources.
+
+	// TODO: consider something more fine-grained than a catch-all "PlatformOperatorError" reason.
+	//   There's a non-negligible difference between "PO is explicitly failing installation" and "PO is not yet installed"
```
tylerslaton, line 97: Maybe we could define these two conditions as reasons and then bubble them up via `util.InspectPlatformOperators()` return? That way we can set that reason inside this `WithAvailable()` call and 
have a less generic "PlatformOperatorError" reason.

timflannagan:
Looks like some of the e2e tests need to be updated too.

/lgtm cancel

timflannagan:
```
{Timed out after 60.001s.
Expected
    <string>: All platform operators are in a successful state
to contain substring
    <string>: All platform operators in a successful state failed /go/src/github.com/openshift/platform-operators/test/e2e/aggregated_clusteroperator_test.go:111
```

Looks like we're missing a couple more test cases :upside_down_face:.

joelanford:
🤮 this is what I get for trying to fix these without actually looking too closely at the failures.

joelanford:
Meta-question: Is there a process I'm not aware of to merge non-bug/non-feature PRs? The label requirements seem to imply one or the other.

timflannagan:
We're doing the TE slides so manually adding the px-approved label after a slack conversation in the #tmp-platform-operators channel.

/label px-approved

Holding, as I think we need QE approval give we're modifying the expected status condition reason/messages/etc. and it's unclear whether that will break any automated testing.

/hold

timflannagan:
This should also be a no-op for docs. Manually adding that label.

/label docs-approved

timflannagan:
Superseding this in #75 given the merge conflicts.

/close

PR #41, opened 2022-09-07
feat: Implement the CPOM component

timflannagan:
# Overview

Implements the CPOM component described in the [phase 0 platform operators EP](https://github.com/openshift/enhancements/blob/master/enhancements/olm/platform-operators.md).

## Summary

This component is responsible for reconciling the PlatformOperators API. During phase 0, admins will configure a desired OLM-based operator package name that's present in the redhat-operators catalog source.

During reconciliation, this component is responsible for finding a registry+v1 bundle that satisfies the admin configured package name. In the case of an invalid configuration, e.g. the configured package doesn't exist, that state will be bubbled up to the PlatformOperator resource the manager is reconciling. After "sourcing" a registry+v1 bundle, the reconciliation logic will generate a rukpak BundleDeployment resource under-the-hood, and delegate the management of that registry+v1 bundle's contents to the rukpak stack.

Note: during phase 0, automatic upgrades are unsupported. In the case that the generated BundleDeployment resource already exists, the reconciliation logic will skip running the sourcing logic to avoid performing unnecessary work.

After generating the BundleDeployment resource that will manage the sourced registry+v1 bundle, the manager will be responsible for inspecting that resource's status sub-resource, and bubble up any failure states to the PlatformOperator resource it's currently reconciling. Common examples may include a registry+v1 bundle that cannot be successfully "unpacked" by rukpak's registry+v1 provisioner, or a registry+v1 bundle contains manifests that are invalid and the Kubernetes API server rejects the creation of that manifest.

## Misc. Notes

- This component will only be deployed when the "TechPreviewNoUpgrades" feature set has been enabled on the cluster.
- When enabling this feature set before cluster creation, admins may wish to inject a series of PlatformOperator custom resources into the cluster's Kubernetes manifests that the installer can use to bootstrap the cluster.
- An "invalid" PlatformOperator resource can influence cluster lifecycle events like ClusterOperator resources can. During phase 0, this means that an invalid PlatformOperator can block cluster rollout. This is done by reconciling an "aggregate" ClusterOperator component that acts as a proxy for the underlying health of a PlatformOperator resource to the CVO component.
- Admins can also create/delete/etc. PlatformOperator custom resources after cluster creation.
- Targeting cert-manager, service-binding-operator, and local-storage-operator as phase 0 candidates.

## Known Limitations

- The reconciliation logic doesn't compare the current and desired state of the BundleDeployment resource. This gap allows users how have access to the BundleDeployment API that ability to manually modify that resource.
- The rukpak provisioner implementations don't have the ability to inspect the health/state of the resources that it's managing. This has implications when bubbling up the generated BundleDeployment state to the PlatformOperator resource that owns it. In the case that a registry+v1 bundle contains manifests that can be successfully applied to the cluster, but will fail at runtime (e.g. a Deployment referencing a non-existent image) the result will be a successful status being reflected in an individual PlatformOperator/BundleDeployment resource.
- Admins configuring these PlatformOperator custom resources before cluster creation cannot easily determine the desired package name without leveraging an existing cluster, or relying on downstream documented examples. There's no validation logic that ensures an individually configured PlatformOperator resource will be able to successfully roll out to the cluster.
- There's no logic that guarantees package name uniqueness.
- During phase 0, the POM component will only source configured packages from the redhat-operators catalog source. This is a short term limitation.


timflannagan:
Just creating something for clusterbot.

api/v1alpha1/types.go
```diff
@@ -16,11 +16,44 @@ limitations under the License.
...
 
+import (
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+)
+
 var (
 	TypeApplied = "Applied"
 
 	ReasonSourceFailed    = "SourceFailed"
+	ReasonUnpackPending   = "ApplyPending"
```
joelanford, line 27:
```suggestion
	ReasonUnpackPending   = "UnpackPending"
```

cmd/main.go
```diff
@@ -56,14 +55,18 @@ func init() {
...
+		metricsAddr          string
+		enableLeaderElection bool
+		probeAddr            string
+	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&runLocal, "run-local", false, "Configures whether the CPOM will run in local dev mode.")
```
joelanford, line 69: Non-blocking nit: see https://github.com/openshift/platform-operators/pull/23#discussion_r965184103

api/v1alpha1/types.go
```diff
@@ -16,11 +16,44 @@ limitations under the License.
...
 	ReasonApplyFailed     = "ApplyFailed"
 	ReasonApplySuccessful = "ApplySuccessful"
 	ReasonApplyPending    = "ApplyPending"
 )
+
+const (
+	SourcedBundleAnnotation = "platform.openshift.io/sourced-bundle"
+)
+
```
joelanford, line 36: Is this related to the convo we had about how to gate sourcing if we've already sourced successfully? If so, add a TODO to switch over to gating based on BD presence? Or did you discover an issue with that approach?

controllers/platformoperator_controller.go
```diff
@@ -37,12 +40,14 @@ import (
 	"github.com/openshift/platform-operators/internal/util"
 )
 
+var (
+	ErrSourceFailed = errors.New("failed to run sourcing logic")
```
joelanford, line 44: Nit: Unexport if not meant to be consumed outside this package.
```suggestion
	errSourceFailed = errors.New("failed to run sourcing logic")
```

controllers/platformoperator_controller.go
```diff
@@ -62,47 +67,82 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 
-	desiredBundle, err := r.Sourcer.Source(ctx, po)
+	return r.reconcile(ctx, po)
+}
+
+func (r *PlatformOperatorReconciler) reconcile(ctx context.Context, po *platformv1alpha1.PlatformOperator) (ctrl.Result, error) {
+	bd, err := r.ensureDesiredBundleDeployment(ctx, po)
 	if err != nil {
+		reason := platformtypes.ReasonApplyFailed
+		status := metav1.ConditionUnknown
```
joelanford, line 90: Should status just always be `"False"`? Both reasons seem to indicate that the BD has not been applied.

controllers/platformoperator_controller.go
```diff
@@ -62,47 +67,82 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-			Message: err.Error(),
-		})
-		return ctrl.Result{}, err
+
+	// check whether the generated BundleDeployment are reporting any
+	// failures when attempting to unpack the configured registry+v1
+	// bundle contents, or persisting those unpack contents to the cluster.
+	if failureCond := util.InspectBundleDeployment(ctx, bd.Status.Conditions); failureCond != nil {
+		meta.SetStatusCondition(&po.Status.Conditions, *failureCond)
+		return ctrl.Result{}, nil
```
joelanford, line 109: Not returning an error here... the assumption being that we'll get a watch event in the future from the BD and that'll be the thing that guarantees we re-reconcile/re-inspect the BD?

Just triple checking that there isn't a terminal state edge to this.

controllers/platformoperator_controller.go
```diff
@@ -62,47 +67,82 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	}
 	meta.SetStatusCondition(&po.Status.Conditions, metav1.Condition{
 		Type:    platformtypes.TypeApplied,
 		Status:  metav1.ConditionTrue,
 		Reason:  platformtypes.ReasonApplySuccessful,
-		Message: "Successfully applied the desired olm.bundle content",
+		Message: fmt.Sprintf("Successfully applied the %s BundleDeployment resource", bd.GetName()),
 	})
+	platformtypes.SetActiveBundleDeployment(po, bd.GetName())
```
joelanford, line 117: I forgot we had this field on the PO status. Could this play a role in gating the sourcing logic?

api/v1alpha1/types.go
joelanford: I see we _are_ already gating on BD presence. Is this dead code?

internal/applier/bundleinstance_applier.go
```diff
@@ -1,56 +1,36 @@
...
-	return err
+	bd.Spec = BuildBundleDeployment(image)
+	return bd
 }
 
-// buildBundleDeployment is responsible for taking a name and image to create an embedded BundleDeployment
-func buildBundleDeployment(image string) *rukpakv1alpha1.BundleDeploymentSpec {
-	return &rukpakv1alpha1.BundleDeploymentSpec{
+// BuildBundleDeployment is responsible for taking a name and image to create an embedded BundleDeployment
+func BuildBundleDeployment(image string) rukpakv1alpha1.BundleDeploymentSpec {
```
joelanford, line 27: What's the need to export this?

internal/applier/bundleinstance_applier.go
```diff
@@ -1,56 +1,36 @@
...
+	return rukpakv1alpha1.BundleDeploymentSpec{
 		ProvisionerClassName: plainProvisionerID,
 		// TODO(tflannag): Investigate why the metadata key is empty when this
 		// resource has been created on cluster despite the field being omitempty.
```
joelanford, line 31: Is this still an issue?

internal/util/util.go
```diff
@@ -111,3 +111,45 @@ func inspectPlatformOperator(po platformv1alpha1.PlatformOperator) error {
...
+		}
+	}
+
+	applied := meta.FindStatusCondition(conditions, rukpakv1alpha1.TypeInstalled)
+	if applied == nil {
+		return &metav1.Condition{
+			Type:    platformtypes.TypeApplied,
+			Status:  metav1.ConditionFalse,
+			Reason:  platformtypes.ReasonUnpackPending,
+			Message: "Waiting for the bundle to be unpacked",
```
joelanford, line 143: Should this be more like "Waiting for bundle to be installed" with reason "InstallPending"?

cmd/main.go
timflannagan: I ended up removing this functionality entirely. I think it's still something that's beneficial for QoL but I didn't want to expand the scope of this work.

api/v1alpha1/types.go
timflannagan: Yup, this was dead code. Removed in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39.

controllers/platformoperator_controller.go
timflannagan: Updated in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39.

controllers/platformoperator_controller.go
timflannagan: Updated in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39.

controllers/platformoperator_controller.go
timflannagan: Added a comment in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39. That was my assumption at the time but forgot to add a comment that explained that behavior. I think we're fine unless we miss an event?

internal/applier/bundleinstance_applier.go
timflannagan: Updated in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39.

internal/util/util.go
timflannagan: Nice catch. Updated in https://github.com/openshift/platform-operators/pull/41/commits/28b4316bc8c76cdb4d61f41877fbe6351ac50f39. I also renamed the "*Applied*" constant variables to "*Install*" as that felt more natural.

internal/applier/bundleinstance_applier.go
timflannagan: Yup, I'm still noticing this locally. I haven't dived too deep on why though.

controllers/suite_test.go
```diff
@@ -17,26 +17,34 @@ limitations under the License.
 package controllers
```
timflannagan, line 17: I think I'll likely revert the changes to this testing suite until I have an implementation that can easily mock catalog sourcing to avoid registering OLM's CatalogSource API here. Longer term, most of the e2e tests introduced in this PR can/should be modeled as unit tests.

internal/sourcer/registry_sourcer.go
```diff
@@ -54,15 +54,12 @@ func (cs catalogSource) Source(ctx context.Context, po *platformv1alpha1.Platfor
 }
 
 func (s sources) GetCandidates(ctx context.Context, po *platformv1alpha1.PlatformOperator) (bundles, error) {
+	// TODO(tflannag): This doesn't account for edge case where there are zero sources.
```
timflannagan, line 57: This is still relevant. I think I need to refactor this `sources` type anyways, and add a test that handles the edge case where the redhat-operators catalog is unavailable.

internal/sourcer/sourcer.go
```diff
@@ -21,3 +21,19 @@ func (b Bundle) String() string {
...
+	b *Bundle
+}
+
+func (no *noopSourcer) SetBundle(b *Bundle) {
+	no.b = b
+}
+
+func (no *noopSourcer) Source(context.Context, *platformv1alpha1.PlatformOperator) (*Bundle, error) {
+	return nil, nil
+}
```
timflannagan, line 39: Note: this is dead code right now.

internal/util/util.go
```diff
@@ -111,3 +111,46 @@ func inspectPlatformOperator(po platformv1alpha1.PlatformOperator) error {
 func buildPOFailureMessage(name, reason string) error {
 	return fmt.Errorf("encountered the failing %s platform operator with reason %q", name, reason)
 }
+
+// InspectBundleDeployment is responsible for inspecting an individual BD
+// resource, and verifying whether the referenced Bundle contents has been
+// successfully unpacked and persisted to the cluster.
+// TODO(tflannag): Add a test for this.
```
timflannagan, line 118: This TODO is still relevant.

timflannagan:
I moved some of the changes outside of this PR and into https://github.com/openshift/platform-operators/pull/44. Should make it easier to remove this PR given the overall diff will be smaller.

internal/sourcer/sourcer.go
timflannagan: Removed in a latest force push.

internal/util/util.go
timflannagan: Added a test for this in the latest force push.

timflannagan:
I already rebased this in the #23 PR which has been getting green CI runs. Moving this out of draft.

controllers/suite_test.go
timflannagan: This is done in the latest force pushes.

controllers/platformoperator_controller.go
```diff
@@ -75,34 +79,71 @@ func (r *PlatformOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+func (r *PlatformOperatorReconciler) ensureDesiredBundleDeployment(ctx context.Context, po *platformv1alpha1.PlatformOperator) (*rukpakv1alpha1.BundleDeployment, error) {
+	bd := &rukpakv1alpha1.BundleDeployment{}
+
+	// check whether the underlying BD has already been generated to determine
+	// whether the sourcing logic needs to be run to avoid performing unnecessary
+	// work given upgrades aren't supported during phase 0. Note: this logic
+	// doesn't compare the current and desired status of the BD resource so it's
+	// possible that users/controllers/etc. can modify the generated BD resource.
+	// See https://github.com/openshift/platform-operators/issues/47 for more details.
+	if err := r.Get(ctx, types.NamespacedName{Name: po.GetName()}, bd); err != nil {
```
tylerslaton, line 131: Do we want to ensure that only one PO with a specific [ActiveBundleDeployment](https://github.com/openshift/api/blob/321abac6303f1c152f2de0cf853b5468ac403051/platform/v1alpha1/types.go#L35-L39) for a package in the catalog is present on cluster as well? Right now we're just checking for PO's that have the same name but those aren't necessarily indicative of their content.

internal/applier/bundledeployment_applied.go
```diff
@@ -1,56 +1,36 @@
 package applier
```
tylerslaton, line 1: Unrelated to this specific PR, but should this file instead be called `bundledeployment_applier.go`? It is `bundledeployment_applied.go` right now.

tylerslaton (commented):
Looks really close, had a few comments.

internal/applier/bundledeployment_applied.go
timflannagan: Whoops. Done in https://github.com/openshift/platform-operators/pull/41/commits/50ba86b0bdada6d75eb6be29359d7f0baebdff98 to so the overall diff is still preserved in the first commit.

controllers/platformoperator_controller.go
tylerslaton: I just tested a situation where RukPak attempts to install two bundles that contain the same content and it looks like the underlying helm logic takes over and prevents that from going through. I think for phase 0 we're fine here in that case.

```
- lastTransitionTime: "2022-09-12T14:10:05Z"
    message: 'rendered manifests contain a resource that already exists. Unable to
      continue with install: Namespace "combo" in namespace "" exists and cannot be
      imported into the current release: invalid ownership metadata; annotation validation
      error: key "meta.helm.sh/release-name" must equal "combo-1": current value is
      "combo"'
```

timflannagan:
Merging the required labels after testing locally on a clusterbot cluster. We've gotten green CI runs in #23 which does have the ability to run e2e.

/label px-approved
/label qe-approved
/label docs-approved

timflannagan:
Whoops.

/override ci/prow/e2e-techpreview

PR #43, opened 2022-09-07
feat: refactor inspectPlatformOperator code and test to be more extensible

tylerslaton:
Adding a small change that verifies the message we are putting out for aggregate CO's with `inspectPlatformOperator`. This also adds a new error that is wrappable.

This is a follow-up to https://github.com/openshift/platform-operators/pull/34#pullrequestreview-1098240110

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 
@@ -18,6 +19,8 @@ import (
 	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
 	platformtypes "github.com/openshift/platform-operators/api/v1alpha1"
 )
+
+var ErrPlatformOperatorUnready = errors.New("platform operator unready")
 
 // GetPodNamespace checks whether the controller is running in a Pod vs.
 // being run locally by inspecting the namespace file that gets mounted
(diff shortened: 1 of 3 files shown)


internal/util/util.go
```diff
@@ -21,7 +22,8 @@ import (
 )
 
 var (
-	ShortRequeue = ctrl.Result{RequeueAfter: time.Second * 5}
+	ShortRequeue                 = ctrl.Result{RequeueAfter: time.Second * 5}
+	ErrorUnreadyPlatformOperator = errors.New("encountered an unready platform operator")
```
tylerslaton, line 26: This may be contentious, so let me know. Mainly adding it so that we can do `errors.Is()` in the unit tests.

exdx, line 26: I would simplify this to simply "platform operator unready"

internal/util/util.go
```diff
@@ -100,14 +102,14 @@ func InspectPlatformOperators(POList *platformv1alpha1.PlatformOperatorList) err
...
-		return buildPOFailureMessage(po.GetName(), applied.Reason)
+		return buildPOMessage(po.GetName(), applied.Reason)
 	}
 	return nil
 }
 
-func buildPOFailureMessage(name, reason string) error {
-	return fmt.Errorf("encountered the failing %s platform operator with reason %q", name, reason)
+func buildPOMessage(name, reason string) error {
+	return fmt.Errorf("%w: %s platform operator has a current reason of %v", ErrorUnreadyPlatformOperator, name, reason)
```
exdx, line 114: This seems a little awkward too. Maybe something like `platform operator is %s, reason` is more legible?

tylerslaton: Maybe we can agree on `%s platform operator has %v reason`?

exdx: I'd like @timflannagan to weigh in

PR #44, opened 2022-09-07
noqe: Remove outdated POC code

timflannagan:
Move some of the unrelated changes in #41 outside of that PR. This removes some dead code that won't be used during phase 0. Removes the magiccatalog library as we're only sourcing from the redhat-operators catalog for 4.12, and upgrades aren't supported. Updates the "applied"-related status condition constants to "installed".


test/e2e/platform_operators_test.go
```diff
@@ -135,14 +59,14 @@ var _ = Describe("platform operators controller", func() {
...
+				Eventually(func() string {
 					bi := &rukpakv1alpha1.BundleDeployment{}
 					if err := c.Get(ctx, types.NamespacedName{Name: po.GetName()}, bi); err != nil {
-						return false
+						return ""
 					}
-					return bi.Spec.Template.Spec.Source.Image.Ref == "quay.io/operatorhubio/prometheus:v0.47.0"
-				}).Should(BeTrue())
+					return bi.Spec.Template.Spec.Source.Image.Ref
+				}).Should(ContainSubstring("registry.redhat.io/cert-manager/cert-manager-operator-bundle"))
```
timflannagan, line 69: I favored this approach to avoid test brittleness over time. There's no guarantee that the redhat-operators catalog won't change the highest semver bundle between patch versions, so this should be more resilient to those kinds of changes.

tylerslaton (approved):
/lgtm

PR #50, opened 2022-09-09
hack: Ensure the aggregate CO reports an Available=True condition

timflannagan:
Update the hack/apply-feature-gate.sh script, and ensure the aggregate platform operators ClusterOperator resource reports an available status.

Moving this out of the #23 PR.


timflannagan:
This is a no-op. Adding the required labels.

/label px-approved
/label docs-approved
/label qe-approved

PR #53, opened 2022-09-09
controllers: Set Progressing=False during success states

timflannagan:
Moving this out of the #23 PR. This was needed as the aggregate CO was blocking cluster rollout despite reporting an Available=True condition when no POs had been installed.

### DIFF
--- a/controllers/aggregated_clusteroperator_controller.go
+++ b/controllers/aggregated_clusteroperator_controller.go
@@ -78,7 +78,7 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 	if len(poList.Items) == 0 {
 		// No POs on cluster, everything is fine
 		coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "No POs are present in the cluster", "NoPOsFound")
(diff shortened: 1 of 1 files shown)


timflannagan:

/label px-approved
/label docs-approved
/label qe-approved

controllers/aggregated_clusteroperator_controller.go
```diff
@@ -90,6 +90,7 @@ func (a *AggregatedClusterOperatorReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, nil
 	}
 	coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, "All POs in a successful state", "POsHealthy")
+	coBuilder.WithProgressing(openshiftconfigv1.ConditionFalse, "All POs in a successful state")
```
tylerslaton, line 93: We should probably have a follow-up to turn PO into PlatformOperators. Not blocking though.

tylerslaton (approved):
/lgtm

controllers/aggregated_clusteroperator_controller.go
timflannagan: Yeah, agreed. We should also move these to constant variables so there's less potential for skew between the implementation and test code.

PR #54, opened 2022-09-09
config,manifests: Update the annotations used in the CVO manifests

timflannagan:
Update the CVO manifests, and update the annotations that are used. Removes the ibm cloud cluster profile for the aggregate CO object, excludes the openshift hosted cluster profile from the annotations, and adds a quote around the "TechPreviewNoUpgrade" feature set.

Moving this out of the #23 PR.

### DIFF
--- a/manifests/0000_50_cluster-platform-operator-manager_00-namespace.yaml
+++ b/manifests/0000_50_cluster-platform-operator-manager_00-namespace.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     release.openshift.io/feature-set: TechPreviewNoUpgrade
   labels:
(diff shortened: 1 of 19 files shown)


timflannagan:

/label px-approved
/label docs-approved
/label qe-approved

manifests/0000_50_cluster-platform-operator-manager_00-namespace.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     release.openshift.io/feature-set: TechPreviewNoUpgrade
```
tylerslaton, line 7: Should these have quotes around them now that you added it to the common annotations?

tylerslaton (approved):
Small comment, but otherwise looks good to me

manifests/0000_50_cluster-platform-operator-manager_00-namespace.yaml
timflannagan: Oh that's weird that it didn't propagate to the generated manifests. I think we're fine here though given it's a no-op and we've seen successfully CI runs in #23.

PR #55, opened 2022-09-09
*: Ensure the aggregate CO controller populates the status.Versions field

timflannagan:
Ensure the aggregated platform operator CO resource has a populated status.Versions field. By default, the YAML manifest for this resource has a populated status.Version but the aggregate CO controller logic stomps on this value during reconciliation. Update that reconciliation logic to ensure this value is persisted during events.

### DIFF
--- a/internal/clusteroperator/release.go
+++ b/internal/clusteroperator/release.go
@@ -1,0 +1,16 @@
+package clusteroperator
+
+import "os"
+
+const (
+	defaultVersionValue    = "0.0.1-snapshot"
+	releaseVersionVariable = "RELEASE_VERSION" // OpenShift's env variable for defining the current release
+)
+
+func GetReleaseVariable() string {
+	v := os.Getenv(releaseVersionVariable)
+	if v == "" {
+		return defaultVersionValue
+	}
+	return v
+}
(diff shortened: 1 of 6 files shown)


timflannagan:

/label px-approved
/label docs-approved
/label qe-approved

internal/clusteroperator/release.go
```diff
@@ -0,0 +1,16 @@
...
+
+const (
+	defaultVersionValue    = "0.0.1-snapshot"
+	releaseVersionVariable = "RELEASE_VERSION" // OpenShift's env variable for defining the current release
+)
+
+func GetReleaseVariable() string {
+	v := os.Getenv(releaseVersionVariable)
+	if v == "" {
+		return defaultVersionValue
```
tylerslaton, line 13: I wonder if us not being able to find the `RELEASE_VERSION` env var should cause an error or not like it does in [OLM today](https://github.com/operator-framework/operator-lifecycle-manager/blob/b2086bd2c5eb9b60370869757c51fa7908d2cde0/pkg/controller/operators/openshift/helpers.go#L214). Is there a situation where this value not being present is okay and we can set a default?

tylerslaton (approved):
Not blocking, but I did have a question before merging.

/hold
/lgtm

internal/clusteroperator/release.go
timflannagan: I did this so it's a bit easier to run locally without having to explicitly set the environment variable. This seems sufficient for now, but it's probably worth revisiting this decision over time.

PR #61, opened 2022-09-19
OCPBUGS-1456: Ensure components can pass openshift conformance tests

timflannagan:
See https://issues.redhat.com/browse/OCPBUGS-1456 for more information.

Adds the openshift-platform-operators system namespace and the CRDs for platform operators/rukpak to the related objects array. Ensures the Pods that get deployed by the CVO follow any openshift conventions defined as conformance tests.


config/manager/manager.yaml
```diff
@@ -57,12 +58,7 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
-        # TODO(user): Configure the resources accordingly based on the project requirements.
-        # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
```
timflannagan, line 65: Note: we have conformance tests that verify component pods don't specify limits, and don't get assigned the BestEffort quality of service. Removing the limits resulted in the Burstable QoS being assigned to these pods.

config/manager/manager.yaml
```diff
@@ -30,6 +30,7 @@ spec:
         runAsNonRoot: true
         seccompProfile:
           type: RuntimeDefault
+      priorityClassName: "system-cluster-critical"
```
timflannagan, line 33: Note: just mirroring what's defined for OLM sub-components here.

config/rukpak/core/resources/deployment.yaml
```diff
@@ -65,6 +70,10 @@ spec:
             - "--bundle-ca-file=/etc/pki/tls/tls.crt"
           ports:
             - containerPort: 8080
+          resources:
+            requests:
+              cpu: 15m
+              memory: 100Mi
```
timflannagan, line 76: This is for the core rukpak deployment. I increased the memory request to 100Mi just because we're caching bundle contents in-memory, so we may want to invest in benchmarking this behavior to get a better read on what kind of requests we need to be setting for subsequent phases.

config/clusteroperator/aggregated_clusteroperator.yaml
```diff
@@ -0,0 +1,22 @@
...
+    resource: namespaces
+  - group: platform.openshift.io
+    name: ""
+    resource: platformoperators
+  - group: core.rukpak.io
+    name: ""
+    resource: bundles
+  - group: core.rukpak.io
+    name: ""
+    resource: bundledeployments
```
joelanford, line 22: Prior to GA, I think we should separate out the rukpak stuff from the PO stuff. At the end of the day, PO will depend on rukpak, not provide it.

But this is fine in the mean time.

timflannagan: Yup +1.

joelanford (approved):
/approve

PR #62, opened 2022-09-19
OLM-2760: Introduce the "core" ClusterOperator controller

timflannagan:
Builds on top of #61 and introduces a "core" ClusterOperator resource and a controller that manages that resource. This controller was described in the phase 0 EP.

### DIFF
--- a/config/clusteroperator/core_clusteroperator.yaml
+++ b/config/clusteroperator/core_clusteroperator.yaml
@@ -1,0 +1,22 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: core
+spec: {}
+status:
+  versions:
+  - name: operator
+    version: "0.0.1-snapshot"
+  relatedObjects:
+  - group: ''
+    name: openshift-platform-operators
+    resource: namespaces
+  - group: platform.openshift.io
+    name: ""
+    resource: platformoperators
+  - group: core.rukpak.io
+    name: ""
+    resource: bundles
+  - group: core.rukpak.io
+    name: ""
+    resource: bundledeployments
--- a/manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
+++ b/manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
@@ -1,0 +1,28 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: TechPreviewNoUpgrade
+  name: platform-operators-core
+  namespace: openshift-platform-operators
+spec: {}
+status:
+  relatedObjects:
+  - group: ""
+    name: openshift-platform-operators
+    resource: namespaces
+  - group: platform.openshift.io
+    name: ""
+    resource: platformoperators
+  - group: core.rukpak.io
+    name: ""
+    resource: bundles
+  - group: core.rukpak.io
+    name: ""
+    resource: bundledeployments
+  versions:
+  - name: operator
+    version: 0.0.1-snapshot
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,7 @@ manifests: generate yq kustomize
 	$(MV_TMP_DIR)/v1_service_platform-operators-controller-manager-metrics-service.yaml manifests/0000_50_cluster-platform-operator-manager_02-metricsservice.yaml
 	$(MV_TMP_DIR)/apps_v1_deployment_platform-operators-controller-manager.yaml manifests/0000_50_cluster-platform-operator-manager_06-deployment.yaml
 	$(MV_TMP_DIR)/config.openshift.io_v1_clusteroperator_platform-operators-aggregated.yaml manifests/0000_50_cluster-platform-operator-manager_07-aggregated-clusteroperator.yaml
+	$(MV_TMP_DIR)/config.openshift.io_v1_clusteroperator_platform-operators-core.yaml manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
 
 	@# cluster-platform-operator-manager rbacs
 	rm -f manifests/0000_50_cluster-platform-operator-manager_03_rbac.yaml
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/utils/clock"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -35,6 +36,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+	"github.com/openshift/platform-operators/internal/checker"
 	"github.com/openshift/platform-operators/internal/clusteroperator"
 	"github.com/openshift/platform-operators/internal/controllers"
 	"github.com/openshift/platform-operators/internal/sourcer"
@@ -99,13 +101,24 @@ func main() {
 	}
 	//+kubebuilder:scaffold:builder
 
-	// Add Aggregated CO controller to manager
+	// Add the core and aggregate ClusterOperator controllers to the manager.
 	if err = (&controllers.AggregatedClusterOperatorReconciler{
 		Client:          mgr.GetClient(),
 		ReleaseVersion:  clusteroperator.GetReleaseVariable(),
 		SystemNamespace: util.PodNamespace(systemNamespace),
 	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "AggregatedCO")
+		setupLog.Error(err, "unable to create controller", "controller", "AggregatedClusterOperator")
+		os.Exit(1)
+	}
+	if err = (&controllers.CoreClusterOperatorReconciler{
+		Client:                mgr.GetClient(),
+		Clock:                 clock.RealClock{},
+		Checker:               checker.ListChecker{Client: mgr.GetClient()},
+		AvailabilityThreshold: clusteroperator.DefaultUnavailabilityThreshold,
+		ReleaseVersion:        clusteroperator.GetReleaseVariable(),
+		SystemNamespace:       util.PodNamespace(systemNamespace),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "CoreClusterOperator")
 		os.Exit(1)
 	}
 
--- a/config/clusteroperator/kustomization.yaml
+++ b/config/clusteroperator/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
 - aggregated_clusteroperator.yaml
+- core_clusteroperator.yaml
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	k8s.io/api v0.25.0
 	k8s.io/apimachinery v0.25.0
 	k8s.io/client-go v0.25.0
+	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
 	sigs.k8s.io/controller-runtime v0.12.3
 )
 
@@ -85,7 +86,6 @@ require (
 	k8s.io/component-base v0.24.2 // indirect
 	k8s.io/klog/v2 v2.70.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1 // indirect
-	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed // indirect
 	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -1,0 +1,31 @@
+package checker
+
+import (
+	"context"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+)
+
+type Checker interface {
+	CheckAvailability(context.Context, *configv1.ClusterOperator) bool
+}
+
+type ListChecker struct {
+	client.Client
+}
+
+func (c ListChecker) CheckAvailability(ctx context.Context, _ *configv1.ClusterOperator) bool {
+	poList := &platformv1alpha1.PlatformOperatorList{}
+	return c.List(ctx, poList) == nil
+}
+
+type NoopChecker struct {
+	Available bool
+}
+
+func (c NoopChecker) CheckAvailability(_ context.Context, _ *configv1.ClusterOperator) bool {
+	return c.Available
+}
--- a/internal/clusteroperator/builder.go
+++ b/internal/clusteroperator/builder.go
@@ -8,6 +8,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ReasonAsExpected = "AsExpected"
+)
+
 // From https://github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorstatus/builder.go
 // Note: the clock api-machinery library in that package is now deprecated, so it has been removed here.
 
--- a/internal/clusteroperator/types.go
+++ b/internal/clusteroperator/types.go
@@ -1,6 +1,9 @@
 package clusteroperator
 
+import "time"
+
 const (
-	CoreResourceName      = "platform-operators-core"
-	AggregateResourceName = "platform-operators-aggregated"
+	CoreResourceName               = "platform-operators-core"
+	AggregateResourceName          = "platform-operators-aggregated"
+	DefaultUnavailabilityThreshold = 5 * time.Minute
 )
--- a/internal/clusteroperator/util.go
+++ b/internal/clusteroperator/util.go
@@ -2,6 +2,7 @@ package clusteroperator
 
 import (
 	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // FindStatusCondition finds the conditionType in conditions.
@@ -13,6 +14,15 @@ func FindStatusCondition(conditions []configv1.ClusterOperatorStatusCondition, c
 		}
 	}
 	return nil
+}
+
+func NewClusterOperator(name string) *configv1.ClusterOperator {
+	return &configv1.ClusterOperator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Status: configv1.ClusterOperatorStatus{},
+	}
 }
 
 // SetDefaultStatusConditions adds the default ClusterOperator status conditions to
--- a/internal/clusteroperator/util.go
+++ b/internal/clusteroperator/util.go
@@ -1,0 +1,16 @@
+package clusteroperator
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// FindStatusCondition finds the conditionType in conditions.
+// Note: manually vendored from o/library-go/pkg/config/clusteroperator/v1helpers/status.go.
+func FindStatusCondition(conditions []configv1.ClusterOperatorStatusCondition, conditionType configv1.ClusterStatusConditionType) *configv1.ClusterOperatorStatusCondition {
+	for i := range conditions {
+		if conditions[i].Type == conditionType {
+			return &conditions[i]
+		}
+	}
+	return nil
+}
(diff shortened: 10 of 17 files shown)


config/clusteroperator/core_clusteroperator.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: core
+spec: {}
+status:
+  versions:
+  - name: operator
+    version: "0.0.1-snapshot"
```
joelanford, line 9: What's the story here? Does `0.0.1-snapshot` have any significance? Is there a reason we statically set this as opposed to leaving it empty here and setting in the controller?

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,105 @@
...
+
+		log.Info("core clusteroperator does not exist. recreating it...", "name", coreCOName)
+		core = &openshiftconfigv1.ClusterOperator{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: coreCOName,
+			},
+			Status: openshiftconfigv1.ClusterOperatorStatus{},
+		}
+		return ctrl.Result{}, r.Create(ctx, core)
+	}
```
joelanford, line 76: If the CO is in the payload, is there ever really a chance that our controller would need to re-create it? Is there a CVO doc about this?

If we do need to be able to re-create it ourselves, it looks like we need to add RBAC `create` permission for COs.

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,105 @@
...
+		if err := coWriter.UpdateStatus(ctx, core, coBuilder.GetStatus()); err != nil {
+			log.Error(err, "error updating CO status")
+		}
+	}()
+
+	// Set the default ClusterOperator conditions to Available=True as the POM isn't
+	// managing any operands during phase 0.
+	coBuilder.WithDegraded(openshiftconfigv1.ConditionFalse)
+	coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, fmt.Sprintf("The platform operator manager is available at %s", r.ReleaseVersion), clusteroperator.ReasonAsExpected)
+	coBuilder.WithProgressing(openshiftconfigv1.ConditionFalse, "")
```
joelanford, line 87: Is there some sort of mechanism by which we can know to set `Progressing=true`? (e.g. when the POM or when one of the core controllers managed under the POM umbrella are being upgraded)

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,105 @@
...
+	defer func() {
+		if err := coWriter.UpdateStatus(ctx, core, coBuilder.GetStatus()); err != nil {
+			log.Error(err, "error updating CO status")
+		}
+	}()
+
+	// Set the default ClusterOperator conditions to Available=True as the POM isn't
+	// managing any operands during phase 0.
+	coBuilder.WithDegraded(openshiftconfigv1.ConditionFalse)
+	coBuilder.WithAvailable(openshiftconfigv1.ConditionTrue, fmt.Sprintf("The platform operator manager is available at %s", r.ReleaseVersion), clusteroperator.ReasonAsExpected)
```
joelanford, line 86: Should one or both of Degraded/Available rely on any smoke tests? (e.g. can I list POs without error?)

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,105 @@
...
+	coBuilder.WithRelatedObject("core.rukpak.io", "bundledeployments", "", "")
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *CoreClusterOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&openshiftconfigv1.ClusterOperator{}, builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			return object.GetName() == coreCOName
```
joelanford, line 102: Complete nit, and maybe hard to justify complexity/value tradeoff, but...

I wonder if we should put a label on our COs like `config.openshift.io/component: platform-operators`, so that we can use a label selector on our cache to avoid caching all COs.

manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: TechPreviewNoUpgrade
+  name: platform-operators-core
+  namespace: openshift-platform-operators
```
joelanford, line 10: Should we be putting a namespace in the CO metadata? Aren't these cluster-scoped objects? Is this just an artifact of kustomize namespace injection? I wonder if there's a way to teach kustomize about the scoping of resources so that it doesn't inject namespaces into cluster-scoped objects.

timflannagan: Yup, this is just an artifact of kustomize injecting these fields. This is also a problem with the aggregate CO YAML manifest. AFAIK, it's a no-op when deployed at runtime.

controllers/core_clusteroperator_controller.go
timflannagan: https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md#how-should-i-include-clusteroperator-custom-resource-in-manifests

> It remains a responsibility of the respective operator to properly update (or recreate if deleted) the ClusterOperator Custom Resource

I was reading through that documentation earlier today and spotted that documentation blob. Makes me question whether we should specify the create-only CVO annotation here if we're fully responsible for this resource outside of the initial rollout.

controllers/core_clusteroperator_controller.go
timflannagan: It depends on what smoke tests we think are reasonable. Reading through https://coreos.slack.com/archives/C030W4V9BHN/p1663354445071699 it seems like setting A=T is sufficient during phase 0 but I'm open to ideas.

controllers/core_clusteroperator_controller.go
timflannagan: Are we treating rukpak as an operand for the core ClusterOperator? I'd assume we aren't as the CVO is deploying that sub-component, and talking about this implementation in https://coreos.slack.com/archives/C030W4V9BHN/p1663354445071699 my understanding was that parsing any of the controller deployments would be anti-pattern here.

config/clusteroperator/core_clusteroperator.yaml
timflannagan: Looking at https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md#what-should-be-the-contents-of-clusteroperator-custom-resource-in-manifests, it sounds like there's logic that looks for this status version and does a copy/replace operation with what's present in the release payload. Not too sure whether this is directly needed in the static YAML manifest, but I've seen a bunch of payload components specify this in their CVO manifests directory.

controllers/core_clusteroperator_controller.go
joelanford: > (or recreate if deleted)

Yeah, this seems to contrast with the fact that CVO would recreate it for us if we didn't use `create-only`. Maybe needs some clarification from @wking.

controllers/core_clusteroperator_controller.go
joelanford: Hmm, yeah I agree rukpak is not really an operand (more of a sibling). Having read the CVO docs you linked. I guess the core PO ClusterOperator doesn't really have an operand -- we've explicitly set the operands apart in the aggregated CO.

> An operator reports Progressing when it is rolling out new code, propagating config changes, or otherwise moving from one steady state to another. It should not report progressing when it is reconciling a previously known state. If it is progressing to a new version, it should include the version in the message for the condition like "Moving to v1.0.1".

It sounds like we'd set `Progressing=true` in the core CO when were performing PO upgrades during a cluster upgrade.

> An operator reports a new version when it has rolled out the new version to all of its operands.

And that also sounds like the point at which we'd set back to `Progressing=false`

controllers/core_clusteroperator_controller.go
joelanford: At a minimum, I think we should attempt a list of POs before we say A=T? I'm trying to think through various POM failures. Here's my stab at filling in a table.

| Failure | platform-operators-core | platform-operators-aggregated |
| ------------- | ------------- | --------------- |
| Can't communicate with apiserver | Available=F  | Available=F |
| Don't have permission to list POs  | Available=F  | Available=F |
| Failed to list bundles from catalog source(s) | Available=F | Available=F |
| Failed to list BDs | Available=F | Available=F |
| Failed to locate PO packageName in listed bundles | Available=T | Available=F |
| Don't have permission to create/update BDs for POs | Available=F | Available=F |
| BD for a PO did not progress to `InstallationSucceeded` | Available=T | Available=F |

>  Failed to list bundles from catalog source(s)

This one seems iffy since I recall that catsrcs GRPC connections can be flaky. Seems like we'd want to do retries for a little bit, then go `Degraded=true`, then eventually go `Available=false`
 
Also from the CVO docs:

> An operator shoould not report the Available status condition the first time until they are completely rolled out (or within some reasonable percentage if the component must be installed to all nodes)

This seems _super_ important to make sure we get right when managing PO installations for POs that are declared at cluster install time. I put #40 on hold while we think through this.

manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
joelanford: It is, but it could be confusing to someone looking at this or the contents of the release image. Maybe something to try to follow-up on later.

timflannagan: Works for me 👍

tylerslaton: If we wanted to, we could put our ClusterOperator files into a separate folder and render them with a separate kustomize call. That would make it so that we can avoid this issue with a little bit of a hack around it.

tylerslaton (commented):
Looks like Joe handled a lot of the questions here already so I left some small comments

controllers/core_clusteroperator_controller.go
timflannagan: How many CO resources would we be potentially be caching downstream? Maybe 15-20 depending on the openshift distribution?

joelanford: yeah, that seems like the right order of magnitude.

manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
joelanford: I feel like there's gotta be a kustomize-centric way to handle this. Does `kustomize` add a `.metadata.namespace` to `Namespace` objects?

controllers/core_clusteroperator_controller.go
timflannagan: Just anecdotal experience is that I can manually delete the ClusterOperator resource, and the CVO won't create a new one. I think the safest thing here is to update both ClusterOperator resources and specify the create-only annotation, and lean on our reconciliation logic to ensure this resource gets created by our controllers.

timflannagan: This is also a gap in the current aggregate ClusterOperator controller implementation. It's a simply change to update the reconciliation logic to create the ClusterOperator if it doesn't already exist, and we can uniformly specify the `release.openshift.io/create-only: "true"` annotation.

controllers/core_clusteroperator_controller.go
timflannagan: This doesn't seem unreasonable but does introduce the edge case where the ClusterOperator has been deleted, and we need to ensure we're labeling the new ClusterOperator resource appropriately. I can tackle this in a follow-up but I'm still on the fence for whether this is worth the performance improvements during this phase.

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,105 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logr "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/openshift/platform-operators/internal/clusteroperator"
+)
+
+type CoreClusterOperatorReconciler struct {
```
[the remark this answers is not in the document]
timflannagan: I think the enhancement proposal should be sufficient here but it's possible I'm biased. And it's unclear whether we'll continue to utilize an aggregate ClusterOperator resource in future phases.

controllers/core_clusteroperator_controller.go
timflannagan: And the value of listing the PlatformOperators in the cluster is to ensure that CRD has been registered? We basically only care about whether an error was returned vs. length checking the number of items the list query returned. I think my main concern is whether that opens the door for transient events, e.g. the API server is temporarily down, and whether toggling A=[T,F] is desirable in such a configuration.

I need to really dive more into the ClusterOperator documentation to get a better feel for what's applicable here given we don't explicitly manage any operands with this CO controller. Thoughts @bparees @wking?

controllers/core_clusteroperator_controller.go
timflannagan: Do you think it's worth tinkering with P=[T,F] toggling during this phase given we implicitly don't support upgrades given the tech preview no upgrade feature set? Maybe that feature set only blocks minor version upgrades vs. blocks patch version upgrades.

A trivial implementation may compare the last recorded and current release version information that could be read from the Pod's `RELEASE_IMAGE` environment variable, and compare whether there's any skew.

controllers/core_clusteroperator_controller.go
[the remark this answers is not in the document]
timflannagan: I was thinking about whether we'd want an in-tree docs/ folder for this repository but didn't go through the full thought experiment yet. That work doesn't have to be a prerequisite to establishing an EP paper trail though.

manifests/0000_50_cluster-platform-operator-manager_07-core-clusteroperator.yaml
tylerslaton: Yeah, there is a part of our kustomize file that is telling it to add namespaces to all resources.

https://github.com/openshift/platform-operators/blob/4ecea427cf5302dfcdf4a5af8d28eadebacc2037/config/default/kustomization.yaml#L2

I'm also sure there is a kustomize centric way to tell it "add namespaces to these, but not these resources"

controllers/core_clusteroperator_controller.go
bparees: checking that you can list the api you care about is definitely more "aggressive" than i think most COs bother with.  That doesn't make it wrong, per se.

It accomplishes two things:

1) ensures the api is defined/available
2) ensures the operator itself has RBAC permissions to interact with the api

That said, i don't know that you need to explicitly poll it, more like if the control loop gets an error interacting with the api, then you might go degraded or unavailable.

timflannagan: > This one seems iffy since I recall that catsrcs GRPC connections can be flaky. Seems like we'd want to do retries for a little bit, then go Degraded=true, then eventually go Available=false

How would we enforce this? My instinct is that this is overkill for phase 0, and we should try to avoid relying on the catalogsource gRPC connection unless we're implementing some sort of reference-counting-esq implementation. Does this also imply we'd need to combine the core ClusterOperator controller with the POM to be able to set CO status conditions during reconciliation?

> That said, i don't know that you need to explicitly poll it, more like if the control loop gets an error interacting with the api, then you might go degraded or unavailable.

Okay so in this implementation that checks whether we can successfully list the POs in the cluster, we'd be setting a degraded status vs. an unavailable status?

bparees: > Okay so in this implementation that checks whether we can successfully list the POs in the cluster, we'd be setting a degraded status vs. an unavailable status?

if you can't list them, i'd say you're unavailable.  But you don't want to go unavailable just because you fail once.  If you fail once, you're degraded.  If you continue to fail for (5? 10? 15? 30?) minutes, you're unavailable.

timflannagan: Do we examples of core components handling this kind of logic? It doesn't seem too difficult to implement how long we've been unable to perform a list query to determine unavailability, but it may involve defining some arbitrary toleration (e.g. 10m) for what we decide is unavailable, and may influence behaviors on other openshift distributions.

bparees: i don't know of one offhand (we more commonly use the pattern on setting Alerts).

Honestly 5 mins should be plenty.  If we can't resolve the api resource after 5 minutes, nothing is likely to change that's going to fix it.

Not sure why it would have different implications on another openshift distribution, i can't think of any where we'd expect the behavior to be different.

timflannagan: Briefly thinking about this some more: I wonder whether we need to distinguish between a "list failure threshold" before we mark the ClusterOperator as degraded. During reconciliation, if we fail to successfully perform a list query 3-5 in a row, then we mark the component as degraded. And then we'd also have a 5m threshold that distinguishes between D=T and A=T.

We're definitely introducing some complexity with this approach, so open to suggestions on whether this is a bad idea. I think my preference is still to keep the core ClusterOperator reconciliation loop pretty stupid during phase 0, and revisit this interaction as a GA requirement once we get a better picture of the overall OLM 1.x stack and our CVO integration model.

bparees: absolutely not a phase 0 requirement.  

as for whether we need a threshold before going degraded...  i could see us just keeping it simple and not going degraded(at least not based on this).  Just have a threshold before you go unavailable.  As long as the threshold is 5 minutes or whatever, i think that's fine.  There's not much benefit to knowing a few minutes earlier "hey this thing is degraded"

So i may have talked us in circles, but ultimately i think "hey, i can't reach the apis i need to do my job, and i've been trying for 5 minutes, so i'm reporting myself unavailable" is probably sufficient.

timflannagan: I can take a stab at this 5m threshold approach to determine A=[T,F] when I'm back on normal dev duties next week.

I'm mainly concerned with the initial cluster rollout edge case, and having a controller reconciliation loop that blindly sets A=T could be sufficient given we can assume the controller pod is running. Obviously it's not the most robust solution but helps side step any forward looking QE bug where we roll out a bad component container image during phase 0.

bparees: > having a controller reconciliation loop that blindly sets A=T could be sufficient given we can assume the controller pod is running.

i think that is the right starting point.

timflannagan: > as for whether we need a threshold before going degraded... i could see us just keeping it simple and not going degraded

I was thinking about this some more today: what happens if we're currently expressing A=T, and we fail a single list query? Ideally, we wouldn't set A=F given it could be transient issue. Would it make sense instead to set D=T here, which would update the LastTransistionTime, and keep the A=T and P=T conditions? Any subsequent reconciliations that continue to fail a list query would check whether D=T is already set, and if so, compare the LastTransistionTime to the failure threshold to determine whether we should be setting A=F.

@bparees @joelanford Thoughts? It was unclear to me reading through the states listed in https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md#conditions-and-installupgrade whether this was a supported configuration.

timflannagan: Took a rough stab at this implementation in https://github.com/openshift/platform-operators/pull/62/commits/cdda9b5e7a77dda2965d768942d67b835afc14e8.

bparees: @timflannagan not sure what you mean by "supported configuration" but the algorithm/determination seems reasonable to me.  Of course you also need to clear D=T once you have a successful list, if and only if the `reason` for the D=T is failure to list.  (i.e. if you were degraded for some other reason, then successfully listing doesn't clear the degraded condition)

controllers/core_clusteroperator_controller.go
timflannagan: I'm going to add a TODO comment for revisiting this because I don't think we immediately need it right now.

controllers/core_clusteroperator_controller.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/openshift/platform-operators/internal/checker"
+	"github.com/openshift/platform-operators/internal/clusteroperator"
+)
+
+// TODO(tflannag): Appropriately set the "Progressing" status condition
+// type during cluster upgrade events.
+// FIXME(tflannag): I'm seeing unit test flakes where we're bumping
+// the lastTransistionTime value despite being in the same state as
+// before which is a bug.
```
timflannagan, line 41: Going to see whether https://github.com/openshift/platform-operators/pull/40 helps improve this behavior given it's refactoring that clusteroperator.Builder library that's being leveraged during reconciliation.

controllers/core_clusteroperator_controller.go
timflannagan: Talked about this in another PR and in slack. Expectation is that CVO#840 will remove the need for managing the ClusterOperator directly, so I removed this logic in recent force pushes/commits.

controllers/core_clusteroperator_controller.go
timflannagan: Talked about this in slack and the decision was to add a TODO comment for addressing this in future phases.

timflannagan:
Moved some of the envtest fixes to #72, which should fix any of the recent unit test case failures where we fail to find the underlying etcd executable. When diving into the failures in CI, it looks like we need to update where the executable directory is stored, and ensure we're writing those files to a directory path that's writable in CI (e.g. /tmp).

tylerslaton:
@timflannagan Anything I can do to help move this forward? Looks like we're blocked by a rebase being needed right now.

PR #63, opened 2022-09-21
OLM-2761: Add validation for package uniqueness

tylerslaton:
# Summary
## Context
There are two things at play for this PR:

1. The PlatformOperator's name
2. The PlatformOperator's `spec.package.name`

## Before

Prior to this PR if two PlatformOperators had different names but the same `spec.package.name` then we would allow them to both get created and the underlying RukPak logic would surface up a collision between them. 

## After
With this PR we are now making it so the core reconciliation loop of the PlatformOperator controller detects when an update would result in a duplicate `spec.package.name` being installed. Instead of allowing the collision to occur during installation we instead now surface a clearer error indicating that a collision would occur.

### DIFF
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -1,0 +1,29 @@
+package validate
+
+import (
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+)
+
+func UniquePackage(ctx context.Context, cli client.Client, po *platformv1alpha1.PlatformOperator) error {
+	existingPlatformOperators := &platformv1alpha1.PlatformOperatorList{}
+	if err := cli.List(ctx, existingPlatformOperators); err != nil {
+		return err
+	}
+
+	for _, otherPO := range existingPlatformOperators.Items {
+		if otherPO.Spec.Package.Name == po.Spec.Package.Name {
+			return fmt.Errorf(
+				"%v spec.package.name is not unique, conflicts with PlatformOperator %v",
+				po.Name,
+				otherPO.Name,
+			)
+		}
+	}
+
+	return nil
+}
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -9,18 +9,25 @@ import (
 	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
 )
 
+// UniquePackage checks that the provided PlatformOperator contains a unique spec.package.name
+// when compared to the other PlatformOperators existing on cluster.
 func UniquePackage(ctx context.Context, cli client.Client, po *platformv1alpha1.PlatformOperator) error {
 	existingPlatformOperators := &platformv1alpha1.PlatformOperatorList{}
 	if err := cli.List(ctx, existingPlatformOperators); err != nil {
 		return err
 	}
 
-	for _, otherPO := range existingPlatformOperators.Items {
-		if otherPO.Spec.Package.Name == po.Spec.Package.Name {
+	for _, existingPO := range existingPlatformOperators.Items {
+		// check whether we're processing the same PlatformOperator as the parameter
+		if existingPO.GetName() == po.GetName() {
+			continue
+		}
+
+		if existingPO.Spec.Package.Name == po.Spec.Package.Name {
 			return fmt.Errorf(
 				"%v spec.package.name is not unique, conflicts with PlatformOperator %v",
 				po.Name,
-				otherPO.Name,
+				existingPO.Name,
 			)
 		}
 	}
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -23,6 +23,8 @@ import (
 var (
 	TypeInstalled = "Installed"
 
+	ReasonValidationFailed = "ValidationFailed"
+
 	ReasonSourceFailed  = "SourceFailed"
 	ReasonUnpackPending = "UnpackPending"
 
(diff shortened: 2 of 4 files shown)


internal/validate/validate.go
```diff
@@ -0,0 +1,29 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errNotUniquePackage = errors.New("spec.package.name is not unique")
+)
+
+func UniquePackage(ctx context.Context, cli client.Client, po *platformv1alpha1.PlatformOperator) error {
+	existingPlatformOperators := &platformv1alpha1.PlatformOperatorList{}
+	if err := cli.List(ctx, existingPlatformOperators); !apierrors.IsNotFound(err) {
```
timflannagan, line 18: Can we get a 404 during a list query? Should we just return an error here?

internal/validate/validate.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"errors"
+
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errNotUniquePackage = errors.New("spec.package.name is not unique")
```
timflannagan, line 13: Do we need to perform any errors.Is operations here? Might be too premature. Not too sure if exported/unexpected makes a difference.

tylerslaton: I was going to use it in our E2E tests so I can easily target it but I get what you're saying here. Its probably premature though.

internal/validate/validate.go
tylerslaton: I don't think we can, you're right. I thought maybe if there were not items?

internal/validate/validate.go
timflannagan: Yeah I think we can just blindly process the error returned by this function at the call site without the want/need to parse based on the error type returned.

internal/validate/validate.go
timflannagan: You'd perform a length check to determine whether there were zero items:

```go
if len(existingPlatformOperators.Items) == 0 {
    ...
}
```

But I don't think we'd need to perform any length checking operations here. My guess is that we wouldn't error out if there were no PO items returned by that list query, despite the PO resource being passed as a parameter as a function. We'd just implicitly accept that we have package uniqueness in this case, and we're working with an out-of-date as no error was returned.

timflannagan:
/retitle OLM-2761: Add validation for package uniqueness

timflannagan (commented):
Looks like we're on the right track. There's some unit test failures I didn't investigate but we should get those checks green before the next round of reviews. I mainly had nits, but I'm wondering if we need integration and e2e feedback given our envtest suite is pretty light.

internal/validate/validate.go
```diff
@@ -0,0 +1,31 @@
...
+	platformv1alpha1 "github.com/openshift/api/platform/v1alpha1"
+)
+
+func UniquePackage(ctx context.Context, cli client.Client, po *platformv1alpha1.PlatformOperator) error {
+	existingPlatformOperators := &platformv1alpha1.PlatformOperatorList{}
+	if err := cli.List(ctx, existingPlatformOperators); err != nil {
+		return err
+	}
+
+	for _, otherPO := range existingPlatformOperators.Items {
```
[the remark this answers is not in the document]
tylerslaton: I see your point, I think that something along the line of `listedPO` given the `po` parameter can be in here as well.

internal/validate/validate.go
```diff
@@ -0,0 +1,31 @@
...
+		notSelf := po.Name != otherPO.Name
+		samePackage := po.Spec.Package.Name == otherPO.Spec.Package.Name
+		if notSelf && samePackage {
```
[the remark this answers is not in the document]
tylerslaton: I think this looks good.

timflannagan (approved):
Can we update the description so it's easier for the no-ff teams to apply the requisite labels? Basically what was the behavior before this change vs. after should be sufficient.

internal/validate/validate.go
```diff
@@ -0,0 +1,36 @@
...
+
+// UniquePackage checks that the provided PlatformOperator contains a unique spec.package.name
+// when compared to the other PlatformOperators existing on cluster.
+func UniquePackage(ctx context.Context, cli client.Client, po *platformv1alpha1.PlatformOperator) error {
+	existingPlatformOperators := &platformv1alpha1.PlatformOperatorList{}
+	if err := cli.List(ctx, existingPlatformOperators); err != nil {
+		return err
+	}
+
+	for _, existingPO := range existingPlatformOperators.Items {
```
joelanford, line 20: I'd propose the we remove the item from this slice that has the earliest timestamp (and tie-break on UID). That way if we're reconciling that one, it won't be considered a duplicate and we'll be able to keep reconciling it.

tylerslaton: That makes sense. What about this situation though:

1. A PlatformOperator `x` gets created with package name `a`
2. Another PlatformOperator `y` gets created with package name `b`
3. I update PlatformOperator `x` to use package name `b` to create a duplicate

What should happen in this case? Technically `x` is the earliest timestamp so by this logic it would reconcile properly.

timflannagan:
We're doing the TE slides so manually adding the px-approval. See https://coreos.slack.com/archives/C030W4V9BHN/p1661875975344669 for more information.

/label px-approved

Holding for QE approval and resolution to https://github.com/openshift/platform-operators/pull/63#discussion_r978183831.

/hold

tylerslaton:
Closing this PR as the current logic works well enough until we have a overarching component that will manage this for us better. The jira issue will remain open in the mean time so that we can eventually come back to this when we have a better view of the landscape.

PR #65, opened 2022-09-22
OLM-2772: Update RukPak and PlatformOperator images via kustomize to be their downtream equivalents

tylerslaton:


timflannagan (approved):
Let's make sure that e2e passes before merging this. Adding the required labels as this should be a no-op and the OLM-2772 has the no-* labels specified.

/label qe-approved
/label docs-approved
/label qe-approved

timflannagan:
Holding for green e2e runs.

/hold

timflannagan:
Manually verified that the relevant images are still being correctly scheduled and replaced with the internal CI registry equivalent.

/hold cancel

PR #66, opened 2022-09-22
OLM-2787: Add support for gather testing artifacts when a test case fails

timflannagan:
Closes #56. Adds a bash script that dumps the YAML representation of the PlatformOperator/BundleDeployment/Bundle resources when failed e2e runs.

PR #68, opened 2022-09-27
Improve the visiblity into individual e2e test case failures

timflannagan:
Related to #66 which has the debug information present.

### DIFF
--- a/test/e2e/collect-ci-artifacts.sh
+++ b/test/e2e/collect-ci-artifacts.sh
@@ -9,6 +9,15 @@ set -o errexit
 : "${KUBECTL:=kubectl}"
 
 function ensure_kubectl() {
+    # Check whether we're running in downstream CI environments as the location for the
+    # "oc" binary is located at "/cli/oc" path. This is problematic as the /cli directory
+    # doesn't exist in the $PATH environment variable, which causes issues when running
+    # this script via the exec.Command Golang function.
+    if [[ "$OPENSHIFT_CI" == "true" ]]; then
+        echo "Detected the e2e suite is being run in CI environment. Adding the /cli to \$PATH"
+        export PATH=$PATH:/cli
+    fi
+
     if ! which ${KUBECTL} &> /dev/null; then
         echo "cannot find kubectl binary in \$PATH"
         exit 1
--- a/test/e2e/collect-ci-artifacts.sh
+++ b/test/e2e/collect-ci-artifacts.sh
@@ -6,7 +6,7 @@ set -o errexit
 
 : "${KUBECONFIG:?}"
 : "${ARTIFACT_DIR:?}"
-: "${KUBECTL:=kubectl}"
+: "${KUBECTL:=oc}"
 
 function ensure_kubectl() {
     # Check whether we're running in downstream CI environments as the location for the
--- a/test/e2e/collect-ci-artifacts.sh
+++ b/test/e2e/collect-ci-artifacts.sh
@@ -1,0 +1,43 @@
+#! /bin/bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+: "${KUBECONFIG:?}"
+: "${ARTIFACT_DIR:?}"
+: "${KUBECTL:=kubectl}"
+
+function ensure_kubectl() {
+    if ! which ${KUBECTL} &> /dev/null; then
+        echo "cannot find kubectl binary in \$PATH"
+        exit 1
+    fi
+}
+
+function collect_artifacts() {
+    commands=()
+    commands+=("get co platform-operators-aggregated -o yaml")
+    commands+=("get platformoperators -o yaml")
+    commands+=("get bundledeployments -o yaml")
+    commands+=("get bundles -o yaml")
+
+    echo "Storing the test artifact output in the ${ARTIFACT_DIR} directory"
+    for command in "${commands[@]}"; do
+        echo "Collecting ${command} output..."
+        COMMAND_OUTPUT_FILE=${ARTIFACT_DIR}/${command// /_}
+
+        ${KUBECTL} ${command} >> "${COMMAND_OUTPUT_FILE}"
+    done
+}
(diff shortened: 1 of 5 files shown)


timflannagan:
Holding for CI feedback in #66.

/hold

timflannagan:
These changes only affect the dev testing suite. Adding the required no-ff labels.

/label px-approved
/label docs-approved
/label qe-approved

tylerslaton (approved):
/lgtm

timflannagan:
Removing the hold now that #66 had e2e runs that expressed the desired behavior.

/hold cancel

test/e2e/util.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	// create a dedicated test case directory to avoid overwriting the
+	// testing artifacts gathered by a previous test case failure.
+	testCaseDir := filepath.Join(artifactDir, strings.ReplaceAll(currentTest.LeafNodeText, " ", "-"))
+	if err := os.MkdirAll(testCaseDir, os.ModePerm); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("/bin/bash", "-c", "./collect-ci-artifacts.sh")
```
timflannagan, line 37: Follow-up: avoid hardcoding the path to the script location.

test/e2e/util.go
```diff
@@ -0,0 +1,48 @@
...
+func HandleTestCaseFailure() error {
+	currentTest := ginkgo.CurrentSpecReport()
+	if !currentTest.Failed() {
+		return nil
+	}
+
+	// current test case failed. attempt to collect CI artifacts if the
+	// $ARTIFACT_DIR environment variable has been set. This variable is
+	// always present in downstream CI environments.
+	artifactDir := os.Getenv("ARTIFACT_DIR")
```
timflannagan, line 24: Potential follow-up: There's overlap between how we handle the $ARTIFACT_DIR environment variable in the Makefile invocation vs. how we're handling the value of that shared variable here. It could be worth consolidating this logic into a shared CLI flag, but it's unclear whether that's the correct implementation going forward.

The main problem with this approach is the potential for skew between how ginkgo handles relative paths vs. how we're handling relative paths within this internal testing code. The expectation is that downstream CI would set the value of $ARTIFACT_DIR variable to an absolute path, so we're in the clear there, but there's no guarantee that we're correctly configuring (or enforcing) the want/need for an absolute path when running the e2e suite locally during dev workflows.

I don't think anything I mentioned is blocking, but it's still something worth noting and potentially revisiting in future phase implementations.

test/e2e/collect-ci-artifacts.sh
```diff
@@ -0,0 +1,52 @@
...
+: "${KUBECONFIG:?}"
+: "${ARTIFACT_DIR:?}"
+: "${KUBECTL:=oc}"
+
+function ensure_kubectl() {
+    # Check whether we're running in downstream CI environments as the location for the
+    # "oc" binary is located at "/cli/oc" path. This is problematic as the /cli directory
+    # doesn't exist in the $PATH environment variable, which causes issues when running
+    # this script via the exec.Command Golang function.
+    if [[ "$OPENSHIFT_CI" == "true" ]]; then
```
timflannagan, line 16: Note: We could also perform a `-d` check on whether the /cli directory exists. The OPENSHIFT_CI variable will always be set when running the e2e suite in downstream CI environments, so it feels like we can make a really solid assumption on where the oc binary will live going forward.

This isn't completely future proof in the case this directory location changes, but we should be fine given we're stat(ing) the configured kubectl binary immediately after performing this conditional check. In the future, updating this script and adding a conditional check for validating this directory exists could produce a more robust implementation.

test/e2e/collect-ci-artifacts.sh
```diff
@@ -0,0 +1,52 @@
+#! /bin/bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+: "${KUBECONFIG:?}"
```
joelanford, line 7: Do we need/use the `KUBECONFIG` variable?

tylerslaton:
Any idea as to what's going on in the CI failure here? I took a look and it seems like things failed pretty quickly so I'm not sure if we know its a flake or not.

timflannagan:
There's some flakes in the e2e testing suite, which is why I wanted to introduce these changes. If you dive into the [test case directory that failed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_platform-operators/68/pull-ci-openshift-platform-operators-main-e2e-aws-techpreview-operator/1574952192236851200/artifacts/e2e-aws-techpreview-operator/test/artifacts/), you can see the relevant YAML output of the resources that get reconciled.

I had the hunch that our consistent usage of cert-manager throughout the e2e suite was problematic as we were constantly installing and then performing a cascading deletion of those contents, and the installing those same contents again without giving any buffer for that deletion to happen. It looks like that hunch is being validated looking at the following status of a failed BundleDeployment resource:

```yaml
    - lastTransitionTime: "2022-09-28T03:37:23Z"
      message: 'rendered manifests contain a resource that already exists. Unable
        to continue with install: Namespace "openshift-cert-manager-operator" in namespace
        "" exists and cannot be imported into the current release: invalid ownership
        metadata; annotation validation error: key "meta.helm.sh/release-name" must
        equal "cert-managerzv79l": current value is "cert-managerlwcq9"'
      reason: InstallFailed
      status: "False"
      type: Installed
```

Not too sure whether this is a rukpak bug, or we need to add some buffer to ensure we're not re-installing the same contents that are being deleted. The potential problem with the latter is that we're hiding controller failures.

test/e2e/collect-ci-artifacts.sh
joelanford: Nit: comment to that effect then? Otherwise someone might decide to delete it in the future.

PR #69, opened 2022-09-28
test/e2e: Use foreground cascading deletion

timflannagan:
When deleting any PlatformOperator or BundleDeployment resources, specify the foreground cascading deletion propagation policy in relevant client usage in the e2e suite.

This should cut down on the number of flakes that result in failed installations due to tests re-using the same PlatformOperator package name. The issue here is we don't give the underlying cascading deletion enough time to fully uninstall before we attempt a new installation. And the result here is that the BundleDeployment installation fails as some resources already exist and cannot be successfully adopted.

### DIFF
--- a/test/e2e/aggregated_clusteroperator_test.go
+++ b/test/e2e/aggregated_clusteroperator_test.go
@@ -87,7 +87,7 @@ var _ = Describe("aggregated clusteroperator controller", func() {
 		})
 		AfterEach(func() {
 			Expect(HandleTestCaseFailure()).To(BeNil())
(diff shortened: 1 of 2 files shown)


timflannagan:
Waiting for CI feedback to see whether it smooths out recent flakes.

/hold

Adding the requisite no-ff labels given these changes are dev testing related and don't touch anything user facing

/label px-approved
/label qe-approved
/label docs-approved

timflannagan:
Looks like tests passed. Going to recycle again to see if there's still some flakes that pop up.

/retest

PR #71, opened 2022-09-28
DONOTMERGE: Check whether envtest can be successfully run

timflannagan:
Pulling debugging out of #62 to help reduce the noise for reviewers.


timflannagan:
Looks like that worked. Opened #72 to fixing this on main vs. the custom branch I was working on at the time.

PR #72, opened 2022-09-28
Ensure downstream CI can run envtest successfully

### LINKED ISSUE #70: Envtest cannot successfully start in CI environments
#62 introduces a couple of tests that leverage envtest, which has exposed an issue where the underlying etcd/apiserver/etc. executables cannot successfully start when running the unit testing suite.

timflannagan:
Closes #70.

Updates the Makefile invocation when running the unit tests to ensure we're able to install the requisite envtest executables in a directory path that's writable in CI. Add a barebones testing suite to ensure we're not regressing on envtest capability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,9 +113,10 @@ lint: golangci-lint
 	GOLANGCI_LINT_CACHE=/tmp/golangci-cache $(GOLANGCI_LINT) run
 
 UNIT_TEST_DIRS=$(shell go list ./... | grep -v /test/)
+ENVTEST_OPTS ?= $(if $(OPENSHIFT_CI),--bin-dir=/tmp)
 .PHONY: unit
 unit: generate envtest ## Run unit tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test -count=1 -short $(UNIT_TEST_DIRS)
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) $(ENVTEST_OPTS) -p path)" go test -count=1 -short $(UNIT_TEST_DIRS)
 
 .PHONY: e2e
 e2e: deploy test-e2e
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -113,9 +113,10 @@ lint: golangci-lint
 	GOLANGCI_LINT_CACHE=/tmp/golangci-cache $(GOLANGCI_LINT) run
 
 UNIT_TEST_DIRS=$(shell go list ./... | grep -v /test/)
+ENVTEST_OPTS ?= $(if $(OPENSHIFT_CI),--bin-dir=/tmp)
```
timflannagan, line 116: I realize that OPENSHIFT_CI could be set to "false" as we're not checking the value of that variable. That said, the existence of this variable should be sufficient for now given local environments are likely setting it, and CI is guaranteed to have it specified.

timflannagan:
This doesn't affect anything user facing, and only affects our dev testing suite. Manually the requisite labels.

/label px-approved
/label docs-approved
/label qe-approved

controllers/suite_test.go
```diff
@@ -32,43 +39,68 @@ import (
...
 }
 
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	ctx, cancel = context.WithCancel(context.Background())
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
+		CRDDirectoryPaths:     []string{"../manifests", "../vendor/github.com/openshift/api/config/v1/"},
```
awgreene, line 65: Why did you choose to remove the `filepath.Join` call?

awgreene (approved):
Looks good! Had one question.

controllers/suite_test.go
awgreene: Or rather, why abandon using filepath.Join.

PR #73, opened 2022-09-29
internal: Move the controllers directory to internal

timflannagan:
Aligning with the pattern we have in upstream o-f repositories. 


timflannagan:
Adding the required labels as this is a no-op and doesn't affect anything user facing.

/label px-approved
/label docs-approved
/label qe-approved

PR #74, opened 2022-09-29
Makefile: Bump the golangci-lint tool dependency to v1.48.x

timflannagan:


timflannagan:
Note: I had into some roadblocks trying to install the v1.49.x release due to go 1.19 issues. IIRC, downstream is still on 1.18 so I ended up settling on the v1.48.x version in the meantime.

This is a no-op, and doesn't touch anything user facing.

/label docs-approved
/label qe-approved
/label px-approved

PR #76, opened 2022-09-30
test/e2e: Preserve the existing environment when using exec.Command

timflannagan:
Ensure we're preserving the existing environment before running any exec.Commands. I noticed locally when running the e2e suite that the underlying bash script wasn't able to find 'oc' despite that executable being in my $PATH, and the `which oc` was outputting strange results when run in isolation. Explicitly specifying the existing environment via `os.Environ()` which should include the $PATH variable helped fix this pathing issue.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -37,12 +37,13 @@ func HandleTestCaseFailure() error {
 	cmd := exec.Command("/bin/bash", "-c", "./collect-ci-artifacts.sh")
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	cmd.Env = append(cmd.Env, []string{
+	envVars := []string{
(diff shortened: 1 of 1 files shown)


timflannagan:
Waiting for CI feedback just in case.

/hold

timflannagan:
alright cool, looks like these changes still work in CI.

/hold cancel

tylerslaton:
@timflannagan Do we not care about the two failures that we're seeing for CI here? Changes look good but want to check first.

timflannagan:
Nope I think it's fine to merge this.

tylerslaton (approved):
Cool, in that case this looks good.

/lgtm

timflannagan:
These changes only affect the internal testing suite. Adding the required labels.

/label qe-approved
/label docs-approved
/label px-approved

PR #80, opened 2023-01-24
Update owners file

awgreene:


OWNERS
```diff
@@ -1,10 +1,13 @@
 approvers:
 - bparees
+- awgreene
+- perdasilva
 - joelanford
 - perdasilva
```
bparees, line 6: i don't think we need Per twice :)

maybe alpha-sort the list to help avoid that problem in the future.

OWNERS
```diff
@@ -1,10 +1,13 @@
...
-- timflannagan
 reviewers:
-- exdx
-- tylerslaton
-- timflannagan
+- dtfranz
+- awgreene
+- perdasilva
+- ankitathomas
+- tmshort
```
bparees, line 12: alpha sort

OWNERS
bparees: (though if we had two of Per, that would certainly be awesome)

awgreene: Ha! Yes it would!

PR #92, opened 2023-10-12
OCPBUGS-20511: Bump golang.org/x/net to v0.17.0

awgreene:


joelanford:
I added a commit to this PR to clear up the unit test failure. See #94 for the original PR I made. I added this commit here because we'll need it as part of the backport.

PR #93, opened 2023-10-13
bump golangci-lint to 1.51.0

joelanford:
We need to bump golangci-lint to at least 1.50 to be able to bump to go 1.20.


joelanford:
/hold

We might want #95 instead.

PR #94, opened 2023-10-14
pin setup-envtest version to match controller-runtime

joelanford:
Using `latest` for setup-envtest is a moving target, which is problematic when using a pinned go verison. This commit pins setup-envtest to the commit aligned with controller-runtime v0.14.4 (b9940ed)


joelanford:
I added this commit to #92 because it will be required for backporting.

PR #95, opened 2023-10-14
switch to bingo for dependency management (and bump golangci-lint@v1.51.0)

joelanford:

PR #64, opened 2022-09-22
OLM-2763: Remove the PlatformOperators CRD cluster profile hacks

timflannagan:
Bumps the o/api dependency to pull in the changes in https://github.com/openshift/api/pull/1282, and removes the in-tree Makefile hacks that injected the requisite cluster profile annotations.

PR #103, opened 2023-10-26
Makefile: Drop yq prerequisite from 'manifests' target

wking:
3dd23a9b68 (#64) removed the final consumer. Subsequently 39662d60ca (#95) removed some other related targets and the `YQ` variable definition.  But this `manifests` prereq snuck through, until this commit, where I'm removing it.


ncdc (approved):

ncdc:
Extremely unlikely this caused e2e to fail, but let's try 1 more time

/retest

PR #108, opened 2024-01-26
NO-ISSUE: Update to use rukpakv1alpha2 apis

kevinrizza:
Includes controller runtime bump changes


cmd/main.go
```diff
@@ -79,8 +79,6 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
```
ncdc, line 83: Did you mean to remove these?

internal/applier/bundledeployment_applier.go
```diff
@@ -24,21 +24,29 @@ func NewBundleDeployment(po *platformv1alpha1.PlatformOperator, image string) *r
...
-					Type: rukpakv1alpha1.SourceTypeImage,
-					Image: &rukpakv1alpha1.ImageSource{
-						Ref: image,
+		Source: rukpakv1alpha2.BundleSource{
+			Type: rukpakv1alpha2.SourceTypeImage,
+			Image: &rukpakv1alpha2.ImageSource{
+				Ref: image,
+			},
+		},
+		/*
```
ncdc, line 38: I would delete this

internal/applier/bundledeployment_applier.go
```diff
@@ -24,21 +24,29 @@ func NewBundleDeployment(po *platformv1alpha1.PlatformOperator, image string) *r
 }
 
 // buildBundleDeployment is responsible for taking a name and image to create an embedded BundleDeployment
-func buildBundleDeployment(image string) rukpakv1alpha1.BundleDeploymentSpec {
-	return rukpakv1alpha1.BundleDeploymentSpec{
+func buildBundleDeployment(image string) rukpakv1alpha2.BundleDeploymentSpec {
+	return rukpakv1alpha2.BundleDeploymentSpec{
 		ProvisionerClassName: plainProvisionerID,
```
varshaprasad96, line 29: Looks like ProvisionerClassName should be `registryProvisionerID`. The bundle provisioner id would go in here.

cmd/main.go
kevinrizza: Yeah.

https://github.com/kubernetes-sigs/controller-runtime/pull/2407

Controller runtime reimplemented the metrics server configuration, and given the state of this project I didn't think it was worth reconfiguring so I just removed the metrics configuration rather than reimplementing it. It wouldn't be a ton of effort to do, but I think we want to try to move toward getting rid of this thing entirely once we start swapping out rukpak.

internal/applier/bundledeployment_applier.go
kevinrizza: I actually think platform operators was specifically designed not to work with registryv1 bundles. Did we lose the plain provisioner in the latest revision?

varshaprasad96: Plain provisioner does exist. I noticed BundleSpec.provisionerClassName to be registry. Which is what is supposed to be specified here. Basically `spec.provisionerClassName` is alpha2 should be equal to `template.spec.provisionerClassName`

kevinrizza: I tested this a little bit and you were correct. We actually don't have support for configuring the provisioner in PO and the only test examples we have support for are registryv1, so the right thing to migrate here is to use the registryv1 provisioner. I've updated accordingly.

varshaprasad96 (approved):
/lgtm
/approve

ncdc:
AWS flake
/retest

PR #113, opened 2024-05-02
OPRUN-3337: Remove platform-operators from the OCP payload

joelanford:
Platform Operators is not currently a priority for the Operator Framework group, and its continued presence and dependence on Rukpak is forcing us to spend time keeping Platform Operators up to date. In order to remain focused on delivering OLMv1, we are removing Platform Operators from OCP. 

It is only enabled under the `TechPreviewNoUpgrade` feature gate, so its removal is not expected to have an impact on production systems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ RUN make build
 FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
 
 COPY manifests /manifests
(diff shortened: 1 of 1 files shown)


joelanford:
I guess there are no origin tests. I will create a 4.16 Jira ticket when I get a chance, and then link it here to satisfy the valid reference checker.

